PR #15, opened 2020-01-30
Add note about ODBC header when installing on Ubuntu

dirkvanrensburg:
I installed adapter in a Docker based on Ubuntu 18.04. Installation failed because of missing ODBC headers which pyodbc depends on. See:

https://github.com/mkleehammer/pyodbc/issues/441

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,9 @@ Easiest install is to use pip:
 
     pip install dbt-sqlserver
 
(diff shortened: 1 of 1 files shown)


mikaelene:
Thanks for the pull request! 🙏

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #32, opened 2020-06-30
necessary changes with synapse to date

swanderz:
opening this PR just to track the changes we've been making to `dbt-sqlserver` to get it working with Synapse. We're halfway through a two-day hackathon. We still a lot of work left to do though. @mikaelene one question -- any strong opinion for how the `sqlserver` and `synapse` adapaters co-exist? as one package with two adapters? or two packages with one adapter each? I'd be interested to hear your thoughts.


mikaelene:
Hi, interesting with a synapse driver! I have no real experience with synapse. But I think there is a lot of difference between azure dw and sql server. If the code won’t be to cluttered by adding support to synapse in dbt-sqlserver I will be happy to do so. What do you think? Is the difference still big enough to keep them separate? One way to see it is like the Postgres and redshift adapters. But the changes may be bigger between those dbs.

mikaelene:
@swanderz how does your adapter work for you? I see that it is not a lot of changes compared to dbt-sqlserver. The ordinary indexing part is not a big deal for synapse I would say. Don't you have some clustering of tables etc to add to synapse?

swanderz:
@mikaelene yeah so my first goal was to get all the integration tests running, which just happened today. you're right that still outstanding are the following pieces:
- some configs for the `DISTRIBUTION` options (CCI, roundrobin, replicated) that I'd love to make available to the `dbt_project.yml`,
- check to make sure the strategies are working properly
- all the ActiveDirectory authentication (which I hope to steal from the [`dbt-azuredw`](https://github.com/1A-Auto/dbt-azuredw/tree/master/dbt) , and potentially
- bringing in some of the external table creation from this repo (https://github.com/fishtown-analytics/dbt-external-tables)

let's table the one package vs. two discussion for a little bit.

swanderz:
closing bc this is broken given `dbt-synapse` is no longer a fork

PR #34, opened 2020-06-30
drop macro sqlserver__drop_relation()

swanderz:
my first thought was to make `sqlserver__drop_relation` just a wrapper for `sqlserver__drop_relation_script`, but then I looked and found it isn't ever used...

### DIFF
--- a/dbt/include/sqlserver/macros/adapters.sql
+++ b/dbt/include/sqlserver/macros/adapters.sql
@@ -39,12 +39,6 @@
   {% endcall %}
 {% endmacro %}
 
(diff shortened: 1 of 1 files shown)


mikaelene:
Thanks! sqlserver__drop_relation is the "real" drop_relation script that dbt uses when they want to drop a relation. I don't exactly remember why i created the other one. But it was for solving an issue with temp-relations laying around and stop new runs from succeed. I think there was difficult to use the "real" one in the other script. I should look into this again to make it look clearer at least.

PR #35, opened 2020-07-02
Fixes #31. Bounded dbt core requirements for v0.15.

### LINKED ISSUE #31: Incompatible with  dbt>0.15, should update setup.py
Reference [issue ](https://github.com/fishtown-analytics/dbt/issues/2423) describes incorrect duplicate macros when using dbt>0.15 with dbt-sqlserver=0.15.

" One helpful change might be for the maintainer of dbt-sqlserver to change their requirements for the package from >=0.15.0 to something like >=0.15.0, <0.16."

alittlesliceoftom:
Added a comment on Readme and bounded the import to prevent import of 0.17 and issue of incorrectly found duplicates [issue 2423](https://github.com/fishtown-analytics/dbt/issues/2423).

Tested by locally installing inside a venv and got dbt-core 0.15.3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@ Passing all tests in [dbt-integration-tests](https://github.com/fishtown-analyti
 Only supports dbt 0.14 and newer!
 - For dbt 0.14.x use dbt-sqlserver 0.14.x
 - For dbt 0.15.x use dbt-sqlserver 0.15.x
+- dbt 0.16.x is unsupported
+- dbt 0.17.x is unsupported  - development in progress
 
 Easiest install is to use pip:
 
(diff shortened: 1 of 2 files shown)


mikaelene:
Thank you @alittlesliceoftom! This is really a required change that makes my life a little easier when not having to explain on slack why it won't work in 0.17 :-)

alittlesliceoftom:
No problem!

Thanks for merging my first open source PR! 😀🎉

alittlesliceoftom:
Hey @mikaelene , 
I just tried installing this as was setting up on some vms. I still had the same original issue. I guess this needs the code being wrapped up in a release? 
I'm not sure on the process for this on this repo.  Is it something I can do?

mikaelene:
Hi, It is not released on pypi (pip). That's why you get the same error. It is quite a process to release anything on pypi and I am the only one that can do that for this package. You could install it from the git repo. Something like `pip install git+https://github.com/mikaelene/dbt-sqlserver.git` might do the trick.

PR #43, opened 2020-08-04
fix pyodbc connection string for custom port

vitorbellini:
Hi.

The previous conn_str did not handle ports different from 1433. The pyodbc.connect receive custom port like `server=ip,port` instead of `server=ip;port=port`


swanderz:
great catch! LGTM

mikaelene:
I agree,  great catch! Merging this now.

PR #29, opened 2020-06-12
V0.17.0

mikaelene:

PR #44, opened 2020-08-26
V0.17.0 -> V0.18.0 changes

swanderz:
just making this so we can track what's changed...

### DIFF
--- a/dbt/include/sqlserver/macros/materializations/snapshot/strategies.sql
+++ b/dbt/include/sqlserver/macros/materializations/snapshot/strategies.sql
@@ -4,7 +4,7 @@
     {% endfor %}), 2)
 {% endmacro %}
 
-{% macro sqlserver__snapshot_check_strategy(node, snapshotted_rel, current_rel, config, target_exists) %}
+{% macro snapshot_check_strategy(node, snapshotted_rel, current_rel, config, target_exists) %}
     {% set check_cols_config = config['check_cols'] %}
     {% set primary_key = config['unique_key'] %}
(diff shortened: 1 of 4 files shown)


dbt/include/sqlserver/macros/materializations/snapshot/strategies.sql
```diff
@@ -4,7 +4,7 @@
     {% endfor %}), 2)
 {% endmacro %}
 
-{% macro sqlserver__snapshot_check_strategy(node, snapshotted_rel, current_rel, config, target_exists) %}
+{% macro snapshot_check_strategy(node, snapshotted_rel, current_rel, config, target_exists) %}
```
swanderz, line 7: why did this go back to `snapshot_check_strategy()`? whereas in  #29  adding back the `sqlserver__` prefix? I'm confused...

mikaelene: Hi, 

I am not 100 sure why it sometimes needs to be named sqlserver__ and sometimes not. But when trying to develop the adapter for 0.16 I got the error that there already was a macro named i.e. `make_relation` and that there could not be two with the same name. As per instructions I renamed all to sqlserver__. When developing for 0.18 the `sqlserver__snapshot_check_strategy` was not picked up and I got the instruction to just name it `snapshot_check_strategy`. See https://getdbt.slack.com/archives/C50NEBJGG/p1598361666015200 for more info.

swanderz:
only opened this as a test

PR #46, opened 2020-08-27
V0.18.0

swanderz:
Upgrade adapter to be compatible with `dbt` version `0.18.0`


dbt/include/sqlserver/macros/materializations/snapshot/snapshot.sql
```diff
@@ -1,206 +1,4 @@
...
-    {% call statement() %}
-      alter table {{ relation }} add "{{ column.name }}" {{ column.data_type }};
-    {% endcall %}
-  {% endfor %}
-{% endmacro %}
\ No newline at end of file
+{% macro sqlserver__post_snapshot(staging_relation) %}
+  -- Clean up the snapshot temp table
+  {% do drop_relation(staging_relation) %}
+{% endmacro %}
```
swanderz, line 4: my understanding is that in removing all these custom snapshots macros, that the default macros are now compatible with `sqlserver`?

mikaelene: Yes! Most of them are :-).

swanderz: ok awesome. this definitely makes things simpler for @nandanhegde15 and I with `dbt-synapse`. Right now `MERGE` is in only private preview for Synapse, so we have to hack our own version of the `snapshot_merge_sql()` macro...

mikaelene:
v0.18.0 seems to work fine so I am merging this to be the new master!

PR #51, opened 2020-09-10
added support for named instance

alangsbo:
Referring https://github.com/mikaelene/dbt-sqlserver/issues/50

If sql server host name contains a backslash - pointing at a named instance connection to server fails

This is because if a named instance is specified, the port number must be omitted. This is fixed by checking if the specified host name contains a "\" and in that case the port number is omitted


mikaelene:
Thanks for the PR @alangsbo !

PR #53, opened 2020-09-17
add support for Azure SQL by adding Active Directory authentication methods

swanderz:
rehash of #49 but to master branch now. @nandanhegde15 will add some more documentation and test this PR does not break the existing on-prem SQL Server connection methods

### DIFF
--- a/dbt/adapters/sqlserver/connections.py
+++ b/dbt/adapters/sqlserver/connections.py
@@ -142,7 +142,7 @@ class SQLServerConnectionManager(SQLConnectionManager):
                 app_id = getattr(credentials, 'AppId', None)
                 app_secret = getattr(credentials, 'AppSecret', None)
 
-            elif getattr(credentials, 'windows_login', False)
+            elif getattr(credentials, 'windows_login', False):
                 con_str.append(f"trusted_connection=yes")
             elif type_auth == 'sql':
                 con_str.append("Authentication=SqlPassword")
--- a/dbt/adapters/sqlserver/connections.py
+++ b/dbt/adapters/sqlserver/connections.py
@@ -22,11 +22,12 @@ def create_token(tenant_id, client_id, client_secret):
     os.environ['AZURE_CLIENT_ID'] = client_id
     os.environ['AZURE_CLIENT_SECRET'] = client_secret
 
-    token = DefaultAzureCredential().get_token('https://database.windows.net//.default')
+    token = DefaultAzureCredential().get_token(
+        'https://database.windows.net//.default')
     # convert to byte string interspersed with the 1-byte
     # TODO decide which is cleaner?
     # exptoken=b''.join([bytes({i})+bytes(1) for i in bytes(token.token, "UTF-8")])
-    exptoken = bytes(1).join([bytes(i, "UTF-8") for i in  token.token])+bytes(1)
+    exptoken = bytes(1).join([bytes(i, "UTF-8") for i in token.token])+bytes(1)
     # make c object with bytestring length prefix
     tokenstruct = struct.pack("=i", len(exptoken)) + exptoken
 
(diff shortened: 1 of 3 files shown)


dbt/adapters/sqlserver/connections.py
```diff
@@ -40,11 +63,17 @@ def type(self):
     def _connection_keys(self):
         # return an iterator of keys to pretty-print in 'dbt debug'
         # raise NotImplementedError
-        return 'server', 'database', 'schema', 'port', 'UID', 'windows_login'
+        if self.windows_login is True:
+            self.authentication = "Windows Login"
+        
+    
+        return 'server', 'database', 'schema', 'port', 'UID', \
+                 'authentication', 'encrypt'
```
NandanHegde15, line 71: So instead of displaying `windows_login` as True in `dbt debug` output , this changes would display `authentication: Windows Login`
We could have added windows auth as an Authentication value within the profiles config but
to have backward compatibility for Windows Auth with the config property : `windows_login ` otherwise `dbt debug` window would display `authentication=sql`

NandanHegde15:
> rehash of #49 but to master branch now. @NandanHegde15 will add some more documentation and test this PR does not break the existing on-prem SQL Server connection methods

I have tested out both the Windows auth and SQL auth  part for an On Prem SQL server with this branch `Connections.py`
image

dbt/adapters/sqlserver/connections.py
```diff
@@ -97,16 +126,55 @@ def open(cls, connection):
...
+            elif type_auth == 'sql':
+                con_str.append("Authentication=SqlPassword")
+                con_str.append(f"UID={{{credentials.UID}}}")
+                con_str.append(f"PWD={{{credentials.PWD}}}")
+
+            if not getattr(credentials, 'encrypt', False):
+                con_str.append(f"Encrypt={credentials.encrypt}")
 
             con_str_concat = ';'.join(con_str)
             logger.debug(f'Using connection string: {con_str_concat}')
```
NandanHegde15, line 160: This code displays the credentials in the `dbt logs` without any encryption. So we would prefer this line be commented out and can be used only during debugging purposes.
This fix can be made in a separate PR but wanted to highlight you this scenario.

swanderz:
@mikaelene this PR is all that is needed to enable Azure SQL, which would pretty cool! Like @NandanHegde15 said, we can authenticate and found all the `dbt` commands to be working.

dbt/adapters/sqlserver/connections.py
```diff
@@ -97,16 +126,55 @@ def open(cls, connection):
...
                 con_str.append(f"trusted_connection=yes")
+            elif type_auth == 'sql':
+                con_str.append("Authentication=SqlPassword")
+                con_str.append(f"UID={{{credentials.UID}}}")
+                con_str.append(f"PWD={{{credentials.PWD}}}")
+
+            if not getattr(credentials, 'encrypt', False):
+                con_str.append(f"Encrypt={credentials.encrypt}")
 
             con_str_concat = ';'.join(con_str)
```
NandanHegde15, line 159: This code displays the credentials in the dbt logs without any encryption. So we would prefer this line be commented out and can be used only during debugging purposes.
This fix can be made in a separate PR but wanted to highlight you this scenario.

dbt/adapters/sqlserver/connections.py
NandanHegde15, line 160: This code displays the credentials in the dbt logs without any encryption. So we would prefer this line be commented out and can be used only during debugging purposes.
This fix can be made in a separate PR but wanted to highlight you this scenario.

mikaelene (approved):
This looks good @NandanHegde15 ! I'll merge it to master and see if I can try it out before pushing to pipi. Feel free to submit at PR for the credentials debug log as well so we solve that security hole.

PR #55, opened 2020-09-24
Update connections.py

NandanHegde15:
1) Added ClientID to be displayed during `dbt debug`  for `Auth=ServicePrincipal`
2) No longer displaying the Creds in `dbt logs` by replacing the actual creds with ***


dbt/adapters/sqlserver/connections.py
```diff
@@ -68,7 +68,7 @@ def _connection_keys(self):
         
     
         return 'server', 'database', 'schema', 'port', 'UID', \
-                 'authentication', 'encrypt'
+               'client_id','authentication', 'encrypt'
```
NandanHegde15, line 71: ClientId was not displayed in `dbt debug` window for `Auth=ServicePrincipal`
image

dbt/adapters/sqlserver/connections.py
```diff
@@ -157,7 +157,18 @@ def open(cls, connection):
...
+            for i, elem in enumerate(con_str):
+                if 'pwd=' in elem.lower():
+                    index.append(i)
+                    
+            if len(index) !=0 :
+                con_str[index[0]]="PWD=***"
+
+            con_str_display = ';'.join(con_str)
+            
+            logger.debug(f'Using connection string: {con_str_display}')
```
NandanHegde15, line 171: Replacing the actual creds with *** in `dbt logs` as users might be using this log file for analysis like what all auth types are being used and how frequently.
So rather than commenting the part, replaced the creds with ***
image

PR #56, opened 2020-10-05
Implement get_columns_in_query macro to fix snapshots with check_cols='all'

qed-:
Fix for snapshots failing if check_cols='all' is set in config, due to default implementation of get_columns_in_query not being valid MS SQL.


mikaelene:
Thank you for this! I didn't know that.

PR #62, opened 2020-11-20
Ci cd take2

mikaelene:
Add the adapter tests

### DIFF
--- a/dbt/adapters/sqlserver/connections.py
+++ b/dbt/adapters/sqlserver/connections.py
@@ -49,7 +49,8 @@ class SQLServerCredentials(Credentials):
     # "sql", "ActiveDirectoryPassword" or "ActiveDirectoryInteractive", or
     # "ServicePrincipal"
     authentication: Optional[str] = "sql"
-    encrypt: Optional[str] = "yes"
+    encrypt: Optional[str] = 'false'
+    trust_cert: Optional[str] = 'false'
 
     _ALIASES = {
         "user": "UID",
@@ -61,6 +62,7 @@ class SQLServerCredentials(Credentials):
         "auth": "authentication",
         "app_id": "client_id",
         "app_secret": "client_secret",
+        "TrustServerCertificate": "trust_cert",
     }
 
     @property
--- a/dbt/adapters/sqlserver/connections.py
+++ b/dbt/adapters/sqlserver/connections.py
@@ -166,7 +166,7 @@ class SQLServerConnectionManager(SQLConnectionManager):
             elif getattr(credentials, "windows_login", False):
                 con_str.append(f"trusted_connection=yes")
             elif type_auth == "sql":
-                con_str.append("Authentication=SqlPassword")
+                #con_str.append("Authentication=SqlPassword")
                 con_str.append(f"UID={{{credentials.UID}}}")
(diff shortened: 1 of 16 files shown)


mikaelene:
The new settings for connections breaks old way of setting up profile. Needs to be fixed before merge. Otherwise the tests are actually running quite good! 

Thanks @swanderz

dbt/adapters/sqlserver/connections.py
```diff
@@ -49,7 +49,8 @@ class SQLServerCredentials(Credentials):
     # "sql", "ActiveDirectoryPassword" or "ActiveDirectoryInteractive", or
     # "ServicePrincipal"
     authentication: Optional[str] = "sql"
-    encrypt: Optional[str] = "yes"
+    encrypt: Optional[bool] = False
```
swanderz, line 52: by "breaks old way of setting up profile" do you mean that I've toggled Encrypt to now be `False` by default instead of `True`?

swanderz: @mikaelene

mikaelene: Something like that. I removed the cert settings stuff from my profile and it stopped working. Just have to make sure the pr doesn’t break the adapter for all users 😀

mikaelene:
@swanderz  I managed to isolate the problem with standard connections to on-prem SQL Server. Looks like when adding Authentcation=SqlPassword to the connections string, some settings around encryption and certificate are changed to a value that doesn't work out of the box. By removing that setting it all works. The question is if it still works with Azure sql with user pass?

swanderz:
EUREKA! You're incredible @mikaelene for finding this. At first glace, I thought "there's no way that can be it", but you're totally right. Finally my lifelong dream to become an expert in Microsoft ODBC Connection String Keywords has been fulfilled. 😝  [This docs page](https://docs.microsoft.com/en-us/sql/connect/odbc/using-azure-active-directory?WT.mc_id=DP-MVP-5003930) has some juicy snippets.



> ### [Example Connection Strings](https://docs.microsoft.com/en-us/sql/connect/odbc/using-azure-active-directory?WT.mc_id=DP-MVP-5003930#example-connection-strings)
> **SQL Server Authentication - legacy syntax.**
> Server certificate is not validated, and encryption is used only if the server enforces it.
> The username/password is passed in the connection string.
> `server=Server;database=Database;UID=UserName;PWD=Password;`
> 
> **SQL Authentication - new syntax.**
> The client requests encryption (the default value of `Encrypt` is true) and the server certificate gets validated, regardless of the encryption setting (unless `TrustServerCertificate` is set to true).
> The username/password is passed in the connection string.
> `server=Server;database=Database;UID=UserName;PWD=Password;Authentication=SqlPassword;`

also this farther down the page
> **Notes**
> To connect using a SQL Server account username and password, you may now use the new `SqlPassword` option, which is recommended especially for Azure SQL since this option enables more secure connection defaults.

swanderz:
@mikaelene everything's good on our end! whenever you're ready

mikaelene:
This looks great! I imagine that snapshot-check did work before. Can it be that it do work but the test have some issues verifying it? Anyway, lets merge this this to master, do some test runs on actual dbt-projects and then publish to pip :-)

Issue #65, opened 2020-12-07
increase CI test coverage for connection target types


swanderz:
give the weird behavior we saw with connection strings and `SqlPassword=True`. @NandanHegde15 had the idea to have a separate Circle CI job would be verify that `dbt debug` works on all target configurations that are commonly used. We can either add this to #62 or in a new PR after #62 is merged.

The ones I can think of:
- [ ] `trusted_connection=True`
- [ ] `windows_integrated=True`
- [ ] when `Encrypt` and `TrustServerCertificate` aren't specified
- [ ] the various combos of`Encrypt`, `TrustServerCertificate`, and `SqlPassword`
- [ ] Active Directory
    - [ ] Password
    - [ ] Integrated (trickier as it would require a VM and an AD user in the same RG as the Azure SQL server)
    - [ ] Interactive

mikaelene:
That would be really great. It would really help improve stability on releases to be confident that this doesn't break. I think after #62 is merged. To make the merges smaller.

swanderz:
@jtcohen6 could you help us think how to best set up running `dbt debug` against all permutations of target configs? for example, we want to make sure that the `connections.py` logic works whether:
- the values are `true` or `false`, and if
- the params aren't specified.
```yaml
encrypt: true
trust_cert: true
```

we'd also need to do the same with all the AAD auth methods...

jtcohen6:
@swanderz I don't think `dbt debug` returns nonzero exit codes, but you could do it just as well by executing a `dbt compile` command with each permutation.

I think you could do that outside of the pytest framework to start: create a `sample.profiles.yml` with a whole range of targets. In your Circle steps, copy that profile and execute `dbt compile --target {target}` for each.

mikaelene:
Is anyone working on this? I would really like to push a new version to pypi. But then we need to make sure that all functionality we are shipping are working as expected...

swanderz:
working on this today!

PR #71, opened 2020-12-29
Add AzureCliCredential to login with CLI

JCZuurmond:
This PR adds the option to use the authentication of the Azure CLI.

See [this closed PR](https://github.com/dbt-msft/dbt-synapse/pull/35) for context. IMHO the CLI login is the best default: you can login with your credentials wo storing them in plain text, you can login with a service principal (again wo storing credentials in plain text) and you can login with an interactive browser session. See [here](https://github.com/dbt-msft/dbt-synapse/pull/35#issuecomment-750402191) for all possible login options of `azure.identitiy`.


dbt/adapters/sqlserver/connections.py
```diff
@@ -49,8 +34,8 @@ class SQLServerCredentials(Credentials):
     # "sql", "ActiveDirectoryPassword" or "ActiveDirectoryInteractive", or
     # "ServicePrincipal"
     authentication: Optional[str] = "sql"
-    encrypt: Optional[bool] = False
-    trust_cert: Optional[bool] = False
+    encrypt: Optional[bool] = True
```
JCZuurmond, line 37: This code comes from [dbt-synapse](https://github.com/dbt-msft/dbt-synapse), there `encrypt` and `trust_cert` are set to True.

dbt/adapters/sqlserver/connections.py
```diff
@@ -159,55 +236,53 @@ def open(cls, connection):
...
-            elif type_auth == "ServicePrincipal":
-                app_id = getattr(credentials, "AppId", None)
-                app_secret = getattr(credentials, "AppSecret", None)
-
             elif getattr(credentials, "windows_login", False):
-                con_str.append(f"trusted_connection=yes")
+                con_str.append("trusted_connection=yes")
             elif type_auth == "sql":
-                #con_str.append("Authentication=SqlPassword")
+                con_str.append("Authentication=SqlPassword")
```
JCZuurmond, line 242: Ditto here about the difference with `dbt-synapse`.

dbt/adapters/sqlserver/connections.py
```diff
@@ -159,55 +236,53 @@ def open(cls, connection):
...
                 con_str.append(f"UID={{{credentials.UID}}}")
                 con_str.append(f"PWD={{{credentials.PWD}}}")
 
-            # still confused whether to use "Yes", "yes", "True", or "true"
```
JCZuurmond, line 173: And here

JCZuurmond (commented):
There are some differences because I started with `dbt-synapse` and then moved the code here so that it can be used in `dbt-synapse` once the inheritance issue is resolved.

dbt/adapters/sqlserver/connections.py
mikaelene: This was changed to not break connections for all users already using dbt-sqlserver. The `encrypt: Optional[bool] = True` would potentially break connections for people when upgrading dbt-sqlserver.

mikaelene:
I think this is a great add-on to the adapter, but it also makes the testing of different connections even more important.

dbt/adapters/sqlserver/connections.py
JCZuurmond: Do we need to set it back here to the `dbt-sqlserver` settings?

JCZuurmond:
Yes, I do not know how we can test this well. I would help here to limit the options `sql-server` supports. With the CLI token we support multiple log in methods at once:

1. Interactive browser
2. User-password
3. Service principal

swanderz:
for more context, the MSFT ODBC Active Directory connection types were built primarily for on-prem AD, rather than Azure AD and they aren't well-suited for using dbt on the command line. They work just fine for logging into a db w/ SSMS, but it would be best  if we never had to:
- store passwords & secrets in plaintext in `~/profiles.yml` (requirement for `ActiveDirectoryPassword` and  `ActiveDirectoryServicePrincipal`)
- log-in with a pop-up every time you  run  a dbt command  (requirement for `ActiveDirectoryInteractive`)
- enable Active Directory Federation Services -- an outdated service that is way too much work for most users  (requirement for `ActiveDirectoryIntegrated`)

Using `azlogin` is ideal because it handles securely caching credentials. @NandanHegde15 and I were just racking our  brains on how we could  do this when @JCZuurmond opened this PR and saved the day.

@mikaelene, when I'm back from PTO next week, addressing #65 will be the first PR. This PR can come after.

dbt/adapters/sqlserver/connections.py
```diff
@@ -1,36 +1,21 @@
...
 from dbt.adapters.base import Credentials
 from dbt.adapters.sql import SQLConnectionManager
-from azure.identity import DefaultAzureCredential
-
 from dbt.logger import GLOBAL_LOGGER as logger
-
-from dataclasses import dataclass
-from typing import Optional
+from azure.core.credentials import AccessToken
+from azure.identity import AzureCliCredential, DefaultAzureCredential
```
swanderz, line 15: any reason these imports got re-arranged? no strong opinion on this side... I believe I used VSCode's "sort imports" context menu

dbt/adapters/sqlserver/connections.py
swanderz: 🤷  probably? I'm leaning towards publishing a breaking change for existing users to make sure that future users are using security best practices. Microsoft themselves have made `Encrypt=True` to be the defaults as well...

dbt/adapters/sqlserver/connections.py
swanderz: again another thing we should address with #65. Also check out https://github.com/dbt-msft/dbt-sqlserver/pull/62#issuecomment-736183924 for the long history of the `SQLPassword`. what a headache.

dbt/adapters/sqlserver/connections.py
swanderz: generally, I think it would be better if this only a PR that changes  `dbt-sqlserver`'s  `connections.py` to add the az cli options, rather than also introducing syntax from `dbt-synapse` especially because `dbt-synapse`'s `connections.py` will be going away soon anyway.

dbt/adapters/sqlserver/connections.py
JCZuurmond: I'll put them back in the same order and open a separate PR for this. I do not care too much either. The following resources are useful:

According to [PEP8](https://www.python.org/dev/peps/pep-0008/#imports), Python imports should be on separate lines (per library) and the imports are grouped in three blocks:
1. Standard library imports.
2. Related third party imports.
3. Local application/library specific imports.

Next to that, it is common to sort the imports within on of these three group alphabetically - to improve searchability. Finally, [some](https://stackoverflow.com/questions/20762662/whats-the-correct-way-to-sort-python-import-x-and-from-x-import-y-statement) put the `import ...` before `from ... import ...`, or vice versa or mix them.

dbt/adapters/sqlserver/connections.py
JCZuurmond: agreed

dbt/adapters/sqlserver/connections.py
```diff
@@ -159,55 +236,53 @@ def open(cls, connection):
                 elif type_auth == "ActiveDirectoryMsi":
                     raise ValueError("ActiveDirectoryMsi is not supported yet")
 
-            elif type_auth == "ServicePrincipal":
-                app_id = getattr(credentials, "AppId", None)
```
JCZuurmond, line 163: These variables were never used, therefore removed

JCZuurmond (commented):
I'll limit the scope of this PR

dbt/adapters/sqlserver/connections.py
JCZuurmond: To limit the scope of this PR, I have set them back to the original defaults, this can be done in a separate PR.

mikaelene:
@JCZuurmond, than you for limiting the scope of your PRs! Makes it much easier to approve. I have clients using this adapter for on-premise SQL Server only. So as long as that still works as before, I am happy to approve all PRs around how o connect to azure sql. As long as it's not to complicated. My only concern here is that it requires installation of the az cli client? But that's maybe not an issue?

JCZuurmond:
@mikaelene : As of now, this PR only adds an option for when `type_auth` is `CLI`, so nothing should break for your clients. The `AzureCliCredential.get_token` methods will raise an error when the Azure cli can not be invoked. We could explicitly add to the README that the Azure CLI is required to be installed, it is implied already.

.circleci/config.yml
```diff
@@ -74,13 +69,58 @@ jobs:
...
+          name: install gnupg
+          command: apt-get install gnupg -y
+      - azure-cli/install
+      - run: *prep=connect
+      - run:
+          name: cnxn -- Azure SQL - SQL CRED user+pass
+          command: |
+            cd test/integration
+            dbt compile --target azuresql_sqlcred
+      - azure-cli/login-with-user:
```
JCZuurmond, line 98: To which Azure account is this login? Or is this something that works for testing only?

swanderz: i have a sandbox Azure tenant, that we're using just for CI purposes for all the dbt-msft repos. happy to share some of the creds if you'd like

JCZuurmond: That is alright, I was curious how the integration test with Azure was setup.

README.md
```diff
@@ -173,7 +206,14 @@ on first run after upgrade.
 - Adds support for Azure Active Directory as authentication provider
 
 #### Fixes:
-- Fix deprecation warning
+- Fix for lack of precision in the snapshot check strategy. (#74 and #56 thanks @qed) Previously when executing two check snapshots the same second, there was inconsistent data as a result. This was mostly noted when running the automatic adapter tests.
```
JCZuurmond, line 209: Maybe this was not correctly merged

swanderz:
@mikaelene this is good to go now!

mikaelene (approved):
Great writeup in the readme! Thanks for all the great work you are putting in.

alittlesliceoftom:
Just shouting that I used this today, and it worked perfectly and easily, fantastic work @JCZuurmond , thank you!

This is a big upgrade in terms of usability and security IMO

Issue #72, opened 2020-12-30
Conventions for Python code


JCZuurmond:
Let's decide on which conventions to apply for Python code.

Issue originates from discussion about python import order

I'll put them back in the same order and open a separate PR for this. I do not care too much either. The following resources are useful:

According to [PEP8](https://www.python.org/dev/peps/pep-0008/#imports), Python imports should be on separate lines (per library) and the imports are grouped in three blocks:
1. Standard library imports.
2. Related third party imports.
3. Local application/library specific imports.

Next to that, it is common to sort the imports within on of these three group alphabetically - to improve searchability. Finally, [some](https://stackoverflow.com/questions/20762662/whats-the-correct-way-to-sort-python-import-x-and-from-x-import-y-statement) put the `import ...` before `from ... import ...`, or vice versa or mix them.

_Originally posted by @JCZuurmond in https://github.com/dbt-msft/dbt-sqlserver/pull/71#discussion_r550109445_

swanderz:
I know @mikaelene is partial to `black` and that the `dbt-core` team uses `flake8`. Happy to discuss any changes via a PR with incremental changes.

sdebruyn:
Pre-commit config was added, follows style similar to dbt-core

PR #73, opened 2020-12-30
WIP: Python conventions

JCZuurmond:
PR in response to Issue #72.

In case we decide to use `pre-commit`:
- [ ] Add section to CONTRIBUTING about using `pre-commit`
- [ ] Run `pre-commit` to see that all tests pass

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,30 @@
+name: "Pre commit"
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8]
+    steps:
+     - uses: actions/checkout@v2
+
+     - name: Set up Python ${{ matrix.python-version }}
+       uses: actions/setup-python@v2
+       with:
+         python-version: ${{ matrix.python-version }}
+
+     - name: Install pre-commit
+       shell: bash
+       run: pip3 install pre-commit
+
+     - name: Run pre-commit hooks
+       shell: bash
+       run: pre-commit run --all-file
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,60 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.4.0
+    hooks:
+      - id: trailing-whitespace
+      - id: check-added-large-files
+      - id: check-json
+      - id: check-ast
+      - id: check-merge-conflict
+      - id: check-toml
+      - id: check-yaml
+        args: [--unsafe]
+      - id: debug-statements
+      - id: detect-private-key
+      - id: end-of-file-fixer
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.790
+    hooks:
+      - id: mypy
+  - repo: https://github.com/ambv/black
+    rev: 20.8b1
+    hooks:
+      - id: black
+        name: Run black formatter
+        language_version: python3.8
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: 3.8.4
+    hooks:
+      - id: flake8
+        name: Run flake8 linter
+        additional_dependencies: ["flake8-bugbear==20.1.4", "pep8-naming==0.11.1"]
+  - repo: https://github.com/timothycrosley/isort
+    rev: 5.6.4
+    hooks:
+      - id: isort
+        additional_dependencies: [toml]
+        name: Sort imports using isort
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v2.7.4
+    hooks:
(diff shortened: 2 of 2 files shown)


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,30 @@
+name: "Pre commit"
```
JCZuurmond, line 1: I used a Github workflow for which I had a snippet. This repo uses circleci with which I have no experience we could port it to there.

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,30 @@
...
+     - uses: actions/checkout@v2
+
+     - name: Set up Python ${{ matrix.python-version }}
+       uses: actions/setup-python@v2
+       with:
+         python-version: ${{ matrix.python-version }}
+
+     - name: Install pre-commit
+       shell: bash
+       run: pip3 install pre-commit
```
JCZuurmond, line 26: [Pre-commit](https://pre-commit.com/hooks.html) is a tool to apply certain conventions - not only for Python. You can install it locally with:

```bash
pip3 install pre-commit
pre-commit install
```

Then all checks are applied before you commit to keep them clean.

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,30 @@
...
+
+     - name: Run pre-commit hooks
+       shell: bash
+       run: pre-commit run --all-file
```
JCZuurmond, line 30: This CI checks if all conventions are actually applied

.pre-commit-config.yaml
```diff
@@ -0,0 +1,60 @@
+repos:
```
JCZuurmond, line 1: @swanderz : I copied something from a cookiecutter project we have, I would suggest you go through them and add comments about which one to use.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - id: check-json
+      - id: check-ast
+      - id: check-merge-conflict
+      - id: check-toml
+      - id: check-yaml
+        args: [--unsafe]
+      - id: debug-statements
+      - id: detect-private-key
+      - id: end-of-file-fixer
+  - repo: https://github.com/pre-commit/mirrors-mypy
```
JCZuurmond, line 16: `mypy` is probably a bit too much now, it requires to set type hinting.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,60 @@
...
+    rev: v0.790
+    hooks:
+      - id: mypy
+  - repo: https://github.com/ambv/black
```
JCZuurmond, line 20: @mikaelene : would you like to keep black?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,60 @@
...
+    rev: 20.8b1
+    hooks:
+      - id: black
+        name: Run black formatter
+        language_version: python3.8
+  - repo: https://gitlab.com/pycqa/flake8
```
JCZuurmond, line 26: Default line length in [pep8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) is 79. [Default of black](https://github.com/psf/black) is 88.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,60 @@
...
+    rev: 3.8.4
+    hooks:
+      - id: flake8
+        name: Run flake8 linter
+        additional_dependencies: ["flake8-bugbear==20.1.4", "pep8-naming==0.11.1"]
+  - repo: https://github.com/timothycrosley/isort
```
JCZuurmond, line 32: This sorts the imports. [By default it](https://github.com/PyCQA/isort#controlling-how-isort-sections-from-imports) applies the three sections and puts `import ...` above `from .. import ...`

JCZuurmond (commented):
I added some stuff we use in other projects. Feel free to comment about what to keep/remove/add.

.pre-commit-config.yaml
mikaelene: I like black! :-). The 79 line length in pep8 is a relic from when the screens weren't wider than 80 chars. But I do not care so much. As long as it's not to complicated to contribute because you must know every nitty gritty piece of the CI/CD workflow I am alright.

mikaelene:
@swanderz , you set this CircleCI thing up. Is this merge ok for you?

.pre-commit-config.yaml
JCZuurmond: Agreed. I would suggest to remove the `mypy`, that highers the bar for contributing (a lot). 

`black` makes the code of multiple contributor more consistent in style, it requires the contributor to run black locally if their changes do not adhere to black.

JCZuurmond:
This PR is not ready to be merge yet, hence I changed it to draft. If we decide which `pre-commit` plugins to keep, I'll run it locally to make sure the checks will pass for the current state of the repo

sdebruyn:
Pre-commit config was added and pre-commit CI will be enabled in the coming days, thanks for the work though!

PR #88, opened 2021-01-21
Resolves #84

### LINKED ISSUE #84: Recreating clustered columnstore index fails on sql server 2014
Hi All!

I´ve setup a project using an sql server 2014 db and ran in to an error when running models: "Incorrect syntax near the keyword 'IF'..."

This seems to occur in the macro `sqlserver__create_clustered_columnstore_index` because of the syntax `DROP INDEX IF EXISTS {{relation_name}}.{{cci_name}}` which will not work in sql server 2014 where the sys tables has to be queried instead.

I tried changing the code in my local fork to:
```
if exists (
        select * from
        sys.indexes where name = '{{cci_name}}'
        and object_id=object_id('{{relation_name}}')
    )
  drop index {{full_relation}}.{{cci_name}}
```

And it seems to work fine in my local dev environment

Br 
Anders

alangsbo:
Resolves clustered index issue on sql 2014. Does not affect whether clustered indexes are created or not


mikaelene:
This is a great add on for supporting older versions of SQL Server.

Issue #91, opened 2021-01-26
clean up relation construction


swanderz:
we should try and standardize on a convention of using `{{ relation }}` wherever possible over these variants:
- `{{ relation.schema }}.{{ relation.identifier }}`
- `to_relation.schema ~ '.' ~ to_relation.identifier`

@jtcohen6, how do we disable by default `database` being included when calling `{{ relation }}`? Do we have create a child of the `Relation` class?

https://github.com/dbt-msft/dbt-sqlserver/blob/4b3f7bce48f3ffa8509a6b16bcd55976a973c2c1/dbt/include/sqlserver/macros/adapters.sql#L49

https://github.com/dbt-msft/dbt-sqlserver/blob/4b3f7bce48f3ffa8509a6b16bcd55976a973c2c1/dbt/include/sqlserver/macros/adapters.sql#L111

https://github.com/dbt-msft/dbt-sqlserver/blob/4b3f7bce48f3ffa8509a6b16bcd55976a973c2c1/dbt/include/sqlserver/macros/adapters.sql#L118

https://github.com/dbt-msft/dbt-sqlserver/blob/4b3f7bce48f3ffa8509a6b16bcd55976a973c2c1/dbt/include/sqlserver/macros/adapters.sql#L130

https://github.com/dbt-msft/dbt-sqlserver/blob/4b3f7bce48f3ffa8509a6b16bcd55976a973c2c1/dbt/include/sqlserver/macros/adapters.sql#L152-L154

https://github.com/dbt-msft/dbt-sqlserver/blob/4b3f7bce48f3ffa8509a6b16bcd55976a973c2c1/dbt/include/sqlserver/macros/adapters.sql#L168-L169

jtcohen6:
+1 to using `{{ relation }}` instead of concatenating.

Check out how we did this same thing in `dbt-spark`:
- [SparkIncludePolicy](https://github.com/fishtown-analytics/dbt-spark/blob/2f7c2dcb419b66d463bb83212287bd29280d6964/dbt/adapters/spark/relation.py#L16-L20), which turns off `database`
- That policy is then set within [SparkRelation](https://github.com/fishtown-analytics/dbt-spark/blob/2f7c2dcb419b66d463bb83212287bd29280d6964/dbt/adapters/spark/relation.py#L26)
[...]

PR #74, opened 2021-01-04
Fix for passing the snapshot adapter test. Fixes #61

mikaelene:

PR #92, opened 2021-01-26
clean up relation usage and default behavior

dataders:
## changes
- implements a `relation.py` which sets the relation:
    -  quoting policy to both:
        - use `"` to quote relation parts, and
    - inclusion policy to never include the database name when referencing a relation, just `schema.identifier`
-  fixes: #91 cleaner relation name gen ( `{{ relation }}` instead of `{{ relation.schema }}.{{ relation.identifier }}`)
- use `adapter.drop_relation` which deprecates `sqlserver__drop_relation_script`
- re-enables the `snapshot_strategy_check_cols` test for Azure SQL (should have been re-enabled with #74)


swanderz:
@jtcohen6 the import is failing but I have no idea why, but I'm willing to bet it's just a rookie python typo.

dbt/adapters/sqlserver/__init__.py
```diff
@@ -1,6 +1,7 @@
 from dbt.adapters.sqlserver.connections import SQLServerConnectionManager
 from dbt.adapters.sqlserver.connections import SQLServerCredentials
 from dbt.adapters.sqlserver.impl import SQLServerAdapter
+from dbt.adapters.sqlserver.relation import SQLServerRelation
```
jtcohen6, line 4: Switch the order of these:
```suggestion
from dbt.adapters.sqlserver.relation import SQLServerRelation
from dbt.adapters.sqlserver.impl import SQLServerAdapter
```
Because `SQLServerAdapter` depends on `SQLServerRelation`, you can't initialize the former before the latter. Here's the error I get locally:
```
Encountered an error:
cannot import name 'SQLServerRelation' from partially initialized module 'dbt.adapters.sqlserver' (most likely due to a circular import) (/Users/jerco/dev/product/plugins/msft/dbt-sqlserver/dbt/adapters/sqlserver/__init__.py)
```

swanderz:
something weird is happening with `drop_relation()`. For all 8 steps of the empty sequence and the first 12 steps of the base test sequence, the `drop_relation`'s work as they should in that

#### this jinja-sql
```
if object_id ('{{ relation }}','{{ object_id_type }}') is not null
    begin
    drop {{ relation.type }} {{ relation }}
    end
```

#### compiles to this
```
 if object_id ('dbt_test_azure_sql_210211000737938430161027.view_model__dbt_backup','V') is not null
      begin
      drop view dbt_test_azure_sql_210211000737938430161027.view_model__dbt_backup
      end
```

However, starting with [step 13 of the base test sequence](https://github.com/fishtown-analytics/dbt-adapter-tests/blob/5fcae1efec199117eeadf820b8daa7fb69a6c520/pytest_dbt_adapter/sequences/base.yml#L36-L40) (and continuing throughout many other tests) it fails, because now the compiled `drop_relation` SQL looks like this:

```
if object_id (''dbt_test_azure_sql_210211000737938430161027'.'swappable'','V') is not null
      begin
      drop view 'dbt_test_azure_sql_210211000737938430161027'.'swappable'
      end
```

@jtcohen6, does the `global_project`'s incremental materialization (and others) change the `QuotePolicy` somehow? wild guess -- does the [`{ this }` relation object](https://github.com/fishtown-analytics/dbt/blob/c8f0469a44f84cfc7d48cb6b938c6cd27288feba/core/dbt/context/providers.py#L1263-L1296) inherit the adapter's quoting policy?


`SQLServerQuotePolicy` is currently set to False for parts, so I don't get why the `swappable` view is getting quoted here.



```python
@dataclass
class SQLServerQuotePolicy(Policy):
    database: bool = False
    schema: bool = False
    identifier: bool = False
```

dbt/include/sqlserver/macros/adapters.sql
```diff
@@ -145,18 +132,18 @@
    type='view')-%}
    {%- set temp_view_sql = sql.replace("'", "''") -%}
 
-   {{ sqlserver__drop_relation_script(tmp_relation) }}
+   {{ sqlserver__drop_relation(tmp_relation) }}
```
jtcohen6, line 135: I believe this, and the other instances below, can just be `{{ adapter.drop_relation(tmp_relation) }}`

jtcohen6 (commented):
> @jtcohen6, does the global_project's incremental materialization (and others) change the `QuotePolicy` somehow? wild guess -- does the `{{ this }}` relation object inherit the adapter's quoting policy?

Hm... the `{{ this }}` relation object definitely inherits the adapter's quoting policy. It also doesn't seem like `{{ this }}` is directly involved here. To be clear, `dbt-sqlserver` reimplements the [incremental materialization](https://github.com/dbt-msft/dbt-sqlserver/blob/master/dbt/include/sqlserver/macros/materializations/incremental/incremental.sql#L32), so I'd think that [this line](https://github.com/dbt-msft/dbt-sqlserver/blob/4dba52d7852b53cb63740dbe9baf139e063adc76/dbt/include/sqlserver/macros/materializations/incremental/incremental.sql#L32) is performing the bad `drop` (because `swappable` indeed `exists_not_as_table`, it exists as a view). I'm not sure what's causing it to quote `old_relation`, though. Feels like your best bet is to take this for a spin locally—create a model as a view, switch it to incremental, add a lot of logging between lines 17 and 32 of the `incremental` materialization, and see when exactly it starts being quoted.

While on this subject, it's also worth checking to see if the default `incremental` materialization still needs overriding—the only difference I can tell is explicitly writing [`{{ tmp_relation.include(schema=False, database=False) }}`](https://github.com/dbt-msft/dbt-sqlserver/blob/4dba52d7852b53cb63740dbe9baf139e063adc76/dbt/include/sqlserver/macros/materializations/incremental/incremental.sql#L8), rather than implicitly [with `{{ tmp_relation }}`](https://github.com/fishtown-analytics/dbt/blob/c8f0469a44f84cfc7d48cb6b938c6cd27288feba/core/dbt/include/global_project/macros/materializations/incremental/helpers.sql#L11) as the default does, even though the result (at least on `dbt-postgres`) is exactly the same.

dbt/include/sqlserver/macros/adapters.sql
swanderz: good catch! qq: why `adapter.drop_relation()`  and not just `drop_relation()`? is it to make sure the macro can't be overwritten by a macro in the project that is also  named `drop_relation`?

swanderz:
> it's also worth checking to see if the default `incremental` materialization still needs overriding

@jtcohen6 I've pulled in changes from #102 which drops the overriding of the incremental materialization. Unfortunately, the same issue is still happening, so I'm going to start locally logging/debugging the global project's incremental materialization.

dbt/adapters/sqlserver/relation.py
```diff
@@ -0,0 +1,27 @@
...
+    database: bool = False
+    schema: bool = True
+    identifier: bool = True
+
+
+@dataclass(frozen=True, eq=False, repr=False)
+class SQLServerRelation(BaseRelation):
+    quote_policy: SQLServerQuotePolicy = SQLServerQuotePolicy()
+    include_policy: SQLServerIncludePolicy = SQLServerIncludePolicy()
+    quote_character: str = '"'
```
swanderz, line 27: @mikaelene this is the one thing I'm not sure about. can you forsee any issues using `"` instead of `'` to quote relation parts?

mikaelene (commented):
The code look really good. There is a lot of changes. How well have you tested this? Have you run it on a real dbt-project or just in the CI?

swanderz:
> Have you run it on a real dbt-project or just in the CI?

Great question. I actually just tried running a project that was previously run with (I believe) `dbt-sqlserver==0.19.0`. And we get an error. So we should def table this PR until `0.19.1`, instead of cramming it prematurely into `0.19.0.1`.

I haven't seen this error message before, maybe @jtcohen6 could explain. It seems it has to do with partial matches due to quoting differences? this might constitute a breaking change for users. What I don't understand is why:
- it searches for an unquoted version, but
- finds a quoted version, when
- the quote policy is none, however
- in all my debugging the relation was always showing up double-quoted (despite the False quote policy)

```
Compilation Error in model SAPHRPeopleDim (models/stage/SAPHRPeopleDim.sql)
  When searching for a relation, dbt found an approximate match. Instead of guessing 
  which relation to use, dbt will move on. Please delete "ajs_stg"."SAPHRPeopleDim", or rename it to be less ambiguous.
  Searched for: ajs_stg.saphrpeopledim
  Found: "ajs_stg"."SAPHRPeopleDim"
  
  > in macro materialization_view_default (macros/materializations/view/view.sql)
  > called by model SAPHRPeopleDim (models/stage/SAPHRPeopleDim.sql)
```

jtcohen6:
> I haven't seen this error message before, maybe @jtcohen6 could explain. It seems it has to do with partial matches due to quoting differences? this might constitute a breaking change for users.

Yes, that's right. If dbt finds an existing relation in the database that is _similar to_ a model's relation-to-e, but not exactly the same (different quoting/casing), dbt will raise a compilation error to avoid accidentally overwriting a view/table it doesn't actually own. [Here's the source](https://github.com/fishtown-analytics/dbt/blob/36d1bddc5b3ec7b0f503972d7138e8260693da38/core/dbt/adapters/base/relation.py#L95-L99) for that behavior.

It sounds like you ran this same project previously, using `dbt-sqlserver==0.19.0`, when dbt _was_ quoting relations. So, now that quoting has been turned off, dbt has identified the discrepancy. I agree this is a breaking change and should probably wait until v0.20.0.

> in all my debugging the relation was always showing up double-quoted (despite the False quote poli

This is the part I'm not sure about! If the `quote_policy` is now `False` across the board, dbt should not be quoting relation components. Which brings me to ask: What's the actual reason to turn off quoting on SQLServer? I think I missed this part of the puzzle. The reason we turn it off for Snowflake is because, if an identifier name is quoted and contains non-uppercase characters, selecting it is a nightmare.

swanderz:
> What's the actual reason to turn off quoting on SQLServer?

great question @jtcohen6. Looks like there wasn't a good reason and I just did it to try and solve a bug that I stirred up along the way. It isn't necessary. It is however necessary to change our relation part quote character from a single quote to double.

@mikaelene I tested this on a local project and it works without issue. ready for for your further review/approval.

mikaelene (commented):
This seems sane. But It is a lot of code to review. I get a little to comfortable with the CI tests sometimes. I think I introduced a bug in snapshots. #108 . So I will review this when I have time.

dbt/adapters/sqlserver/relation.py
swanderz: Yeah we did, inexplicably though, the code change when I change it back to single quote. You're right that we should think about this change maybe breaking existing snapshot or incremental tables...

dbt/include/sqlserver/macros/adapters.sql
```diff
@@ -108,19 +95,19 @@
...
   {% call statement('rename_relation') -%}
-    EXEC sp_rename '{{ from_relation.schema }}.{{ from_relation.identifier }}', '{{ to_relation.identifier }}'
+    EXEC sp_rename '{{ from_relation }}', '{{ to_relation.identifier }}'
     IF EXISTS(
     SELECT *
     FROM sys.indexes
-    WHERE name='{{ from_relation.schema }}_{{ from_relation.identifier }}_cci' and object_id = OBJECT_ID('{{ from_relation.schema }}.{{ to_relation.identifier }}'))
-    EXEC sp_rename N'{{ from_relation.schema }}.{{ to_relation.identifier }}.{{ from_relation.schema }}_{{ from_relation.identifier }}_cci', N'{{ from_relation.schema }}_{{ to_relation.identifier }}_cci', N'INDEX'
+    WHERE name='{{ from_relation.schema }}_{{ from_relation.identifier }}_cci' and object_id = OBJECT_ID('{{ from_relation }}'))
+    EXEC sp_rename N'{{ from_relation }}.{{ from_relation.schema }}_{{ from_relation.identifier }}_cci', N'{{ from_relation.schema }}_{{ to_relation.identifier }}_cci', N'INDEX'
```
panasenco, line 110: Why are we prefixing with the entire relation now? That includes double-quoted schema and relation names, which is creating really bizarre output in my run:

```
EXEC sp_rename N'"dbo"."APPT_COMPLETED_MONTHLY__dbt_tmp".dbo_APPT_COMPLETED_MONTHLY__dbt_tmp_cci', N'dbo_APPT_COMPLETED_MONTHLY_cci', N'INDEX'
```

dbt/adapters/sqlserver/relation.py
panasenco: I will definitely have to change this back to include `database=True` by default to make cross-database selects work.

panasenco:
@swanderz I've spent several hours today trying to reimplement database switching (see PR #111) in this branch. However, I've run into a wall with the function [`adapter.drop_relation`](https://docs.getdbt.com/reference/dbt-jinja-functions/adapter/#drop_relation). The macro `create_table_as` does the following steps:

1. Drop the temporary view if it exists
2. Create the temporary view
3. Drop the temporary table if it exists
3. Insert temporary view contents into the temporary table
4. Drop the temporary view again

Because the temporary view is dropped twice, something screwy happens with the [relation cache](https://docs.getdbt.com/docs/guides/creating-new-materializations#update-the-relation-cache) and the second drop view statement doesn't happen, leaving the temporary view in the database. I'm not certain what exactly is happening because honestly the relation cache is difficult to understand.

My thought is to just drop all uses of `adapter.drop_relation` and to do it with straightforward SQL rather than trying to understand dbt's relation cache. I don't see the point of the cache anyway, it shaves off milliseconds at most... But I want to check in with you before I do that and ask whether that's a direction you're willing to go.

Is it OK to replace all uses of `adapter.drop_relation` with some straightforward macro?

swanderz:
@panasenco -- I very much appreciate you taking the time to try and make your PR with my refactor. TIL, I had no idea about the relation cache.

Totally open to the "straightforward SQL" approach, but any chance you could open up a different PR with the code you were stuck on so we can look at the error you were seeing? If anything, for curiosity's sake.

PR #93, opened 2021-01-27
break up and disptach-ify get_columns_in_relation

swanderz:
### background

Synapse doesn't have a way to get column names from a temp table ([relevant SO question](https://stackoverflow.com/questions/63800841/get-column-names-of-temp-table-in-azure-synapse-dw))

This PR enables simpler logic for working around Synapse's lack of `tempdb.INFORMATION_SCHEMA.COLUMNS`. The logic change would be as simple as implementing `synapse__get_columns_in_relation_temp` in which we call something like what @jtcohen6 suggested:

```sql
{% macro synapse__get_columns_in_relation_temp(relation) -%}
    {% set select_star %} select * from {{ relation }} {% endset %}
    {{ return(get_columns_in_query(select_star)) }}
{% endmacro %}
```

### changes
With this PR, `get_columns_in_relation` will now call either or the following (now-dispatched) macros depending if `relation.identifier.startswith("#")`:
- `get_columns_in_relation_normal` (for non-temp tables/views), or
- `get_columns_in_relation_temp` (for temp tables)

### questions
1. @mikaelene can you think of a situation where we might want to query both `tempdb.INFORMATION_SCHEMA.COLUMNS` and `INFORMATION_SCHEMA.COLUMNS` for a given temp table? or is the `UNION ALL` just a workaround?
2. for organization's sake, perhaps we splinter the `get_columns_*` macros out of `adapters.sql` and into a `columns.sql` file?
3. @jtcohen6 the current error message is failing at [this line of the snapshot materialization](https://github.com/fishtown-analytics/dbt/blob/540c1fd9c6d501911f6a3a3335bae63fc10924f8/core/dbt/include/global_project/macros/materializations/snapshot/snapshot.sql#L237)... I suspect that `sqlserver__get_columns_in_relation_temp` isn't working yet...
4. @jtcohen6 I'm uncertain about discrepancy of return types for two two `get_columns_in_*` macros (see below). Are the other columns attributes (`data_type`, `character_maximum_length`, `numeric_precision`, `numeric_scale`) ever used? My gut says that if  `synapse__get_columns_in_relation_temp` returns a list of strings, then the [snapshot_materialization will break here](https://github.com/fishtown-analytics/dbt/blob/540c1fd9c6d501911f6a3a3335bae63fc10924f8/core/dbt/include/global_project/macros/materializations/snapshot/snapshot.sql#L239-L244).
    1. `get_columns_in_relation()` - a list of `Column` objects, and
    2. `get_columns_in_query()` - a list of strings of column names

### DIFF
--- a/dbt/include/sqlserver/macros/adapters.sql
+++ b/dbt/include/sqlserver/macros/adapters.sql
@@ -176,41 +176,59 @@
   getdate()
 {%- endmacro %}
 
-{% macro sqlserver__get_columns_in_relation(relation) -%}
+{% macro get_columns_in_relation_normal(relation) %}
+  {{ return(adapter.dispatch('get_columns_in_relation_normal')(relation)) }}
+{% endmacro %}
+
+{% macro default__get_columns_in_relation_normal(relation) %}
+  {{ exceptions.raise_not_implemented(
+    'get_columns_in_relation_normal macro not implemented for adapter '+adapter.type()) }}
+{% endmacro %}
+
+{% macro sqlserver__get_columns_in_relation_normal(relation) %}
   {% call statement('get_columns_in_relation', fetch_result=True) %}
-      SELECT
-          column_name,
-          data_type,
-          character_maximum_length,
-          numeric_precision,
-          numeric_scale
-      FROM
-          (select
-              ordinal_position,
-              column_name,
-              data_type,
-              character_maximum_length,
-              numeric_precision,
-              numeric_scale
-          from INFORMATION_SCHEMA.COLUMNS
-          where table_name = '{{ relation.identifier }}'
-            and table_schema = '{{ relation.schema }}'
-          UNION ALL
-          select
-              ordinal_position,
-              column_name collate database_default,
-              data_type collate database_default,
-              character_maximum_length,
-              numeric_precision,
-              numeric_scale
-          from tempdb.INFORMATION_SCHEMA.COLUMNS
-          where table_name like '{{ relation.identifier }}%') cols
-      order by ordinal_position
-
-
+  SELECT
+    column_name,
+    data_type,
+    character_maximum_length,
+    numeric_precision,
+    numeric_scale
+  FROM INFORMATION_SCHEMA.COLUMNS
+  WHERE table_name = '{{ relation.identifier }}'
+    AND table_schema = '{{ relation.schema }}'
   {% endcall %}
-  {% set table = load_result('get_columns_in_relation').table %}
-  {{ return(sql_convert_columns_in_relation(table)) }}
+  {{ return(sql_convert_columns_in_relation(load_result('get_columns_in_relation').table)) }}
+{% endmacro %}
+
+{% macro get_columns_in_relation_temp(relation) %}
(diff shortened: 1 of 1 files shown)


jtcohen6:
@swanderz I'm conflicted here. I get that by doing the separation and dispatch here, `dbt-synapse` can use `sqlserver__get_columns_in_relation_normal` as is, and just reimplement `synapse__get_columns_in_relation_temp`.

At the same time, I think this PR is adding a lot of complexity to `dbt-sqlserver`, on account of a quirk that exists only in Synapse, when:
- the `union all` approach seemed to be working just fine for SQLServer, and is similar to how most databases handle/support this need
- the relevant logic _could_ be encoded entirely within `synapse__get_columns_in_relation`

I agree that the discrepancy of return types between `get_columns_in_relation` and `get_columns_in_query` is something that we'd need to address. Only the `name` value of each `Column` object is strictly necessary for the snapshot materialization here, including the call to `adapter.get_missing_columns` ([source](https://github.com/fishtown-analytics/dbt/blob/46d36cd4123e3ee3cdb293e9555115905efbaf0d/core/dbt/adapters/base/impl.py#L591-L606)). **Of course, if the snapshot really _does_ need to add columns to the target table because of new columns in the source query, we won't have the right data type while doing it.** That feels like a _big_ limitation of temp tables on Synapse today, and [hopefully it won't be that way forever](https://docs.microsoft.com/en-us/answers/questions/84168/idenitfying-the-columns-in-temp-table-in-synapse.html). I'm not sure how much more surgery we can do to handle all the edge cases in the meantime.

For now, I _think_ we could manage a hacky fix in `dbt-synapse` alone that reimplements `get_columns_in_relation` like so:
```sql
{% macro synapse__get_columns_in_relation(relation) -%}
  
  {# temp tables need different handling #}
  {% if relation.identifier.startswith("#") %}
    {% set columns_in_temp = [] %}
    {% set select_star %} select 1 as fun, true as you {% endset %}
    {% if execute %}
      {% set colnames = get_columns_in_query(select_star) %}
      {% for name in colnames %}
        {# should we set a default data type?? #}
        {% set col_obj = api.Column.create(name, '') %}
        {% do columns_in_temp.append(col_obj) %}
      {% endfor %}
    {{ columns_in_temp }}
    {% endif %}
  {% else %}
    {% call statement('get_columns_in_relation', fetch_result=True) %}
      SELECT
          column_name,
          data_type,
          character_maximum_length,
          numeric_precision,
          numeric_scale
      FROM
          (select
              ordinal_position,
              column_name,
              data_type,
              character_maximum_length,
              numeric_precision,
              numeric_scale
          from INFORMATION_SCHEMA.COLUMNS
          where table_name = '{{ relation.identifier }}'
            and table_schema = '{{ relation.schema }}') cols
  {% endcall %}
  {% set table = load_result('get_columns_in_relation').table %}
  {{ return(sql_convert_columns_in_relation(table)) }}
  
  {% endif %}

{% endmacro %}
```

swanderz:
closing as we have simpler solution with https://github.com/dbt-msft/dbt-synapse/pull/42

PR #95, opened 2021-01-28
V0.19.0 release

swanderz:
- bumped version number
- moved changelog to standalone file (it's getting long!)
- deleted those pesky `__init__.py` files bc:
    - they're  not needed
    - I don't remember why I added them
    - now they're gone less clicking in VS Code to open these dirs... 
      image


swanderz:
I thought this would be a quick PR, but now our connection tests are failing for no good reason that I can think of...

mikaelene:
Everything works great when running pytest locally, but not when running with tox. I get this weird error: 
```
ERROR: Cannot install dbt-sqlserver 0.19.0 (from /Users/mikaelene/dev/dbt-sqlserver) because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested dbt-sqlserver 0.19.0 (from /Users/mikaelene/dev/dbt-sqlserver)
    The user requested dbt-sqlserver 0.19.0 (from /Users/mikaelene/dev/dbt-sqlserver)
```

mikaelene (commented):
Looks great. I re-added the __init__.py file and now it works. You probably added the plutil stuff to make it work in circle CI.

mikaelene (approved):

Issue #63, opened 2020-11-25
Add application name to ODBC string


jtcohen6:
Microsoft collects anonymous aggregate statistics about SQLServer + Synapse usage via the "Application Name" property of connections. After we spoke with a Synapse PM the other day, they're curious to know how many people are using dbt with SQLServer + Synapse.

If you're open to it, I believe we can add an `APP` parameter to the ODBC connection string. For reference:

- [SQL Server docs](https://docs.microsoft.com/en-us/sql/relational-databases/native-client/applications/using-connection-string-keywords-with-sql-server-native-client?view=sql-server-ver15#odbc-driver-connection-string-keywords)
- Similar code in `dbt-spark`: we pass a [useragent for ODBC connections to Databricks](https://github.com/fishtown-analytics/dbt-spark/blob/cde01217aa78451fb4d7b906e1c1bec2ec6edf5a/dbt/adapters/spark/connections.py#L352)

swanderz:
Totally doable, IMHO. this might also dovetail with https://github.com/dbt-msft/dbt-synapse/issues/18 in that the `APP` param could depend on which product is specified in the `profiles.yml`?

mikaelene:
we could maybe send "dbt"+ `adapter name from profiles.yml`

jtcohen6:
@mikaelene @swanderz Would you be up for getting this in before the v0.19.0 release? I think sending `'dbt'` + `credentials.type()` makes a lot of sense!

mikaelene:
I can look into that before release

mikaelene:
Solved with #63

PR #98, opened 2021-02-02
Update application name ODBC parameter

jtcohen6:
My bad for directing us to the wrong docs in #63. Instead of `APP` we want `Application Name`. From Synapse PM:

> On the application name parameter, for ODBC, please use “Application Name=dbt-{credentials.type}” (per your example). The key here is the parameter being “Application Name” with a space between Application and Name. If you happen to use JDBC, then you can use applicationName (no space).  

While we're here, I figure it can't hurt to send the `dbt-sqlserver` version as well.

### DIFF
--- a/dbt/adapters/sqlserver/connections.py
+++ b/dbt/adapters/sqlserver/connections.py
@@ -10,6 +10,7 @@ from typing import Callable, Mapping
 import dbt.exceptions
 from dbt.adapters.base import Credentials
 from dbt.adapters.sql import SQLConnectionManager
+from dbt.adapters.sqlserver import __version__
 from dbt.contracts.connection import AdapterResponse
 from azure.core.credentials import AccessToken
 from azure.identity import AzureCliCredential, DefaultAzureCredential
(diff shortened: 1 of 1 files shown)


swanderz:
thanks @jtcohen6! two qqs:
- If I want to include this change as part of dbt-synapse `v0.19.0`, does that mean we should cut a new release of dbt-sqlserver?
- if so, what version should that be? I'm hesitant to go with `v0.19.1` bc it'll likely be a dbt-core version we might need to support. Maybe `v0.19.0.1`? But, IIRC, any addition after the patch number makes PyPI think it is a pre-release.

jtcohen6:
Yes to a new release, and I think `v0.19.0.1` is the right way to go! We released [`dbt-spark==0.18.1.1`](https://pypi.org/project/dbt-spark/0.18.1.1/#history) when we needed a small fix, but still wanted to indicate version parity with `dbt-core==0.18.1`. I believe that PyPi/pip treat a version as pre-release IFF its suffix is `dev`, `a`, `b`, or `rc` ([docs](https://packaging.python.org/guides/distributing-packages-using-setuptools/#pre-release-versioning)).

Issue #100, opened 2021-02-05
make CLI authentication case-insensitive


swanderz:
in  #99, we found out that users assume that `authentication: CLI` can also be given as `authentication: cli` which is not the case.

@JCZuurmond... if you're interested 👀 

Spitball idea, but maybe we add an `.upper()` to the this snippet below. something like:
```python
if type_auth.upper() in AZURE_AUTH_FUNCTIONS.keys()
```

https://github.com/dbt-msft/dbt-sqlserver/blob/0d529de2641cc7e3f84bab47e22a9f4c80629bbd/dbt/adapters/sqlserver/connections.py#L281-L288

JCZuurmond:
@swanderz : That would indeed solve the problem for when authentication type is `cli` instead of `CLI` (or `cLi` for that matter). However it would not make the other authentication types case insensitive.

I have the urge to restructure some code first so that we can add (py)tests. It would make me more confident about changing code without breaking it.

However, if we prefer to do a quick fix first, then that is ok with me.

swanderz:
no urgency to make a hotfix, given that this issue is documented and easily fixed. more tests would be great -- thanks for offering. Happy to add any tests you write to CircleCI.

PR #103, opened 2021-02-06
Add pytest for unit testing

JCZuurmond:
This PR adds pytest for unit testing.

I tried to follow the dbt setup as much as possible. They have a [`unit` test folder](https://github.com/fishtown-analytics/dbt/tree/develop/test/unit) with unit tests. I followed the pytest best practice to mimick the [package structure](https://docs.pytest.org/en/reorganize-docs/new-docs/user/directory_structure.html) as well. 

It is a little awkward since SQL tests and python tests are combined. The set-up expects the usually pytest directory structure in the `test/unit` folder - while the convention is to put those in the `tests/` folder.


mikaelene:
We had a discussion about testing the different ways for connection. Is this to support that? I guess that we need to clarify what is what with regards to the dbt package tests. Have the other adaptors anything like this?

JCZuurmond:
yes, this indeed intended to test the different ways for connection. I want to use it for #100. And more generally it can be used for testing the python code.

`dbt` and `dbt-spark` have such a `test/unit` folder for python testing. I added subfolders to organize the test in the usual pytest way.

@swanderz : Could you have a look to see if the CI will run the python unit tests.

swanderz:
@JCZuurmond looking like we're in the same boat as before w.r.t getting CI working from your fork. i just changed some settings, can you try pushing a new commit to see if my tweaks fixed things?

JCZuurmond:
the CI ran, the checks only show `integration-sqlserver`

JCZuurmond:
CI passes!

tox.ini
```diff
@@ -6,13 +6,17 @@ envlist = unit, flake8, integration-sqlserver, integration-synapse
...
-    -e.
+    -r {toxinidir}/requirements.txt
 
 [testenv:integration-azuresql]
 basepython = python3
 commands = /bin/bash -c '{envpython} -m pytest -v test/integration/azuresql.dbtspec'
 passenv = DBT_AZURESQL_DB DBT_AZURESQL_PWD DBT_AZURESQL_SERVER DBT_AZURESQL_UID
 deps =
-    -r{toxinidir}/requirements.txt
-    -e.
```
JCZuurmond, line 18: @swanderz : Why the `-e` editable flag for testing?

swanderz: @JCZuurmond great question! my gut tells me it isn't needed, but I just copied directly from [dbt-spark's `tox.ini`](https://github.com/fishtown-analytics/dbt-spark/blob/90492319adce4345199a59527fc9787228bb25dc/tox.ini#L21-L28). I'm very much a newbie to `pytest` and `tox`. Feel free to remove it!

swanderz: for more, context, [here's me on SO trying to get help](https://stackoverflow.com/questions/65784102/install-local-package-in-tox-unencumbered-by-previous-install-from-github) solving a bizarre bug we were having. The answer was great, but actually gave me even more questions! Maybe there's something useful in there for you..

swanderz:
@JCZuurmond this PR is much appreciated. I'm certainly a python testing and packaging newbie, so help in this area is great. I had originally assumed that the [`test/unit/` dir of dbt-spark](https://github.com/fishtown-analytics/dbt-spark/tree/master/test/unit) was for testing the Python classes and functions, but it looks as if there's some integration testing in there as well?

I think we're lucky that with the `connection-sqlserver` and `connection-azuresql` CircleCI jobs, we cover most of the of the connection profile differences without having to roll our own `pytest` suite. My original thought is that this PR would serve so we could write unit tests for the python functions you've added, namely:
- `convert_bytes_to_mswindows_byte_string`
- `convert_access_token_to_mswindows_byte_string`
- `get_cli_access_token`, and
- `get_sp_access_token`.

But I'm unsure how the last two could be unit tests without also being integration tests... 🤷 
That said this PR is all about setting the stage for future unit tests.
> I want to use it for #100. And more generally it can be used for testing the python code.



as for this... I have no idea why there is only one check running instead of four.
> the CI ran, the checks only show integration-sqlserver



a few asks -- can you add:
1. pull from master and add a changelog entry?
2. code formatting check for `black` @mikaelene's partial to it over `flake8`

tox.ini
JCZuurmond: Check, I removed it for both. I am new to `tox`, I have quite some experience with python testing and heard about tox. What I "know" is that it allows you to create different environments to do your testing in. It is useful for things like: run the test suite for python 3.7, 3.8 and 3.9 or a windows and linux system.

I had a quick look at the [tox](https://tox.readthedocs.io/en/latest/index.html) website. Tox uses [virtual environments](https://docs.python.org/3/tutorial/venv.html) to install the package in and then run your test suite. The editable flag is not needed since tox will recreate the environment if it detects a change: `If certain aspects of the project change, a re-creation of the environment is automatically triggered.`

Looking at the tox website I think we use it a bit awkwardly here, but I do not want to change too much in one PR. 

Tox will install your package for your inside the virtual environment it creates, it uses the `setup.py` to do so - `python setup.py sdist` to be precise. However, we tell it to skip this step with the `skipsdist = True` flag. But... we in the end still install the package because there is a dot `.` at the end of the requirements file which tells pip to install the package in the current directory - `pip` is by default used to run the commands in `deps` against.

I think it would make more sense to remove the `skipsdist = True` flag and move the requirements dependencies into the `setup.py`. Seeing that the other `dbt` repos do it in this way, there is a benefit to sticking to it tough.

JCZuurmond:
@swanderz : I can shoot in a first PR to for the unit tests. I was thinking to do some [mocking](https://docs.python.org/3/library/unittest.mock.html) to test the azure login methods.

1. I have added a changelog entry
2. I would do that in a different PR. Maybe recommendation would be to go for #73 and use `pre-commit`. But seeing that dbt does not use `pre-commit`, I am ok with not usin it.

This PR is ok to be merged in my opinion!

swanderz:
> as for this... I have no idea why there is only one check running instead of four.
> 
> > the CI ran, the checks only show integration-sqlserver

So deleted a blank line and now all the jobs run... 


More importantly, everything is passing now! LGTM

PR #105, opened 2021-02-11
Clean up tox ini

JCZuurmond:
As discussed in #103, here some suggestion to change the use of `setup.py`, `requirements.txt` and `tox.ini`. I think we should merge #103  first and sync this with that one.


requirements.txt
```diff
@@ -1,9 +1,5 @@
-dbt-core==0.19.0
```
JCZuurmond, line 1: these were duplicate, i.e. already present in `setup.py`

requirements.txt
```diff
@@ -1,9 +1,5 @@
...
-pyodbc>=4.0.27
-azure-identity>=1.4.0
 black~=20.8b1
 pytest-dbt-adapter~=0.4.0
 tox==3.2.0
 flake8>=3.5.0
 certifi==2020.6.20
```
JCZuurmond, line 6: is this used in the package? Or something like `tox` and `flake`, i.e. used when developing the package.

tox.ini
```diff
@@ -1,18 +1,6 @@
 [tox]
-skipsdist = True
-envlist = unit, flake8, integration-sqlserver, integration-synapse
+envlist = py37,py38
```
JCZuurmond, line 2: our test suite is ran for python 3.7 and 3.7

.circleci/config.yml
```diff
@@ -30,7 +30,7 @@ jobs:
           pkg-manager: pip
       - run:
           name: Test adapter on SQL Server against dbt-adapter-tests
-          command: tox -e integration-sqlserver
+          command: tox -- -v test/integration/sqlserver.dbtspec
```
JCZuurmond, line 33: this is how `tox` recommends you to [pass down positional arguments](https://tox.readthedocs.io/en/latest/example/general.html#interactively-passing-positional-arguments) together with {posargs}.

dev_requirements.txt
```diff
@@ -1,9 +1,6 @@
-dbt-core==0.19.0
-pyodbc>=4.0.27
-azure-identity>=1.4.0
 black~=20.8b1
```
JCZuurmond, line 1: I renamed it to `dev_requirements` since it are the packages you need to install for local development. Dbt also has a `dev_requirements.txt`

tox.ini
```diff
@@ -1,18 +1,6 @@
...
-basepython = python3
-commands = /bin/bash -c '{envpython} -m pytest -v test/integration/azuresql.dbtspec'
-passenv = DBT_AZURESQL_DB DBT_AZURESQL_PWD DBT_AZURESQL_SERVER DBT_AZURESQL_UID
-deps =
-    -r{toxinidir}/requirements.txt
-    -e.
\ No newline at end of file
+[testenv]
+commands = pytest {posargs}
+deps = pytest~=6.2.2 pytest-dbt-adapter~=0.4.0
```
JCZuurmond, line 6: these are the test dependencies

dev_requirements.txt
```diff
@@ -1,9 +1,6 @@
...
 pytest-dbt-adapter~=0.4.0
+pytest~=6.2.2
 tox==3.2.0
 flake8>=3.5.0
 certifi==2020.6.20
```
JCZuurmond, line 6: is this used for development or for the package itself?

JCZuurmond:
The CI is failing because it is trying to use some cache, could we force it to not do that @swanderz ?

.circleci/config.yml
```diff
@@ -26,18 +26,14 @@ jobs:
...
   connection-sqlserver:
     <<: *sqlserver
     steps:
       - run: *install-git
       - checkout
-      - python/install-packages:
```
JCZuurmond, line 39: not needed since this is what tox does for you

.circleci/config.yml
```diff
@@ -6,7 +6,9 @@ orbs:
 
 jobs:
   integration-sqlserver: &sqlserver
-    docker: 
+    docker:
+      - image: cimg/python:3.7
```
JCZuurmond, line 10: we need this to run testing for multiple python versions

JCZuurmond:
@swanderz : When I try to install multiple python versions the CI does not work. Could you help?

dev_requirements.txt
```diff
@@ -1,9 +1,6 @@
-dbt-core==0.19.0
-pyodbc>=4.0.27
-azure-identity>=1.4.0
 black~=20.8b1
 pytest-dbt-adapter~=0.4.0
```
JCZuurmond, line 2: fix version

dev_requirements.txt
swanderz: not as far as I can tell. might as well try removing it!
```suggestion
```

setup.py
```diff
@@ -57,8 +57,8 @@ def _dbt_sqlserver_version():
         ]
     },
     install_requires=[
-        'dbt-core~=0.19.0',
+        "dbt-core>=0.19.0",
```
swanderz, line 60: I'm fairly certain that we want to keep the tilde character bc (according to this [SO answer](https://stackoverflow.com/a/39590286/3842610)), it is a combination of to a combination of the following clauses:
- `>= 0.19.0` and
- `== 0.19.*`

This way a user will get an error if they try to install `dbt-core==0.20.0` before we cut a compatible release.

swanderz:
> When I try to install multiple python versions the CI does not work. Could you help?

for "unauthorized" issue, there's a setting that should is already enabled  for this repo: `Pass secrets to builds from forked pull requests`. In t[he docs for this feature](https://circleci.com/docs/2.0/oss/#build-pull-requests-from-forked-repositories), it does bring up this scenario. Are you by chance "following" your fork from your personal GitHub account rather than from the dbt-msft org? My view looks like below, with dbt-msft in the top left...
>  Note: If a user submits a pull request to your repository from a fork, but no pipeline is triggered, then the user most likely is following a project fork on their personal account rather than the project itself of CircleCi, causing the jobs to trigger under the user’s personal account and not the organization account. To resolve this issue, have the user unfollow their fork of the project on CircleCI and instead follow the source project. This will trigger their jobs to run under the organization when they submit pull requests.

image


https://circleci.com/docs/2.0/oss/#example-open-source-projects

.circleci/config.yml
```diff
@@ -78,25 +74,21 @@ jobs:
...
       - run:
           name: wake up serverless db
           command: |
             cd test/integration
             dbt debug --target azuresql_sqlcred
       - run:
           name: Test adapter on Azure SQL against dbt-adapter-tests
-          command: tox -e integration-azuresql
+          command: tox -- -v test/integration/azuresql.dbtspec
   connection-azuresql:
```
swanderz, line 85: I think we still need to install the dependencies here. for `connection-azuresql` and `connection-sqlserver` I'm cheating by using not using `tox` or `pytest`, instead, we:
- install the dependencies for `dbt-sqlserver` using the `python/install-packages` CirceCI command
- install `gnupg` (note: this should be added to the docker container)
- install the azure-cli,
- move to the dir of the integration test's `dbt_project.yml`
- call `dbt compile` to debug a connection with:
    - a SQL cred
    - an AAD account, and
    - service principal credential

swanderz (changes requested):
totally on board with this clean up. next steps (as I see them):
- pull in #103 
- get CircleCI jobs passing
- add entry to changelog
- get consensus from Mikael and Jeremy on this new approach

setup.py
JCZuurmond: Sure, that is true. Though, we might miss when there is a new minor version and thus forget to leverage that.

The projects I use as examples - like `[tox](https://github.com/tox-dev/tox/blob/master/setup.cfg)` or `[pre-commit](https://github.com/pre-commit/pre-commit/blob/master/setup.cfg)` - have lower bounded requirement version. Sometimes you see an upper bound for the major release `dbt-core>=0.19.0,<1.0.0`. 

I think the idea is that `dbt-core` should be backwards compatible for minor releases, however, it could of course happen that that is not the case.

swanderz: good point. no strong opinion here. @mikaelene might have one though!

JCZuurmond:
> > When I try to install multiple python versions the CI does not work. Could you help?
> 
> for "unauthorized" issue, there's a setting that should is already enabled for this repo: `Pass secrets to builds from forked pull requests`. In t[he docs for this feature](https://circleci.com/docs/2.0/oss/#build-pull-requests-from-forked-repositories), it does bring up this scenario. Are you by chance "following" your fork from your personal GitHub account rather than from the dbt-msft org? My view looks like below, with dbt-msft in the top left...
> 
> > Note: If a user submits a pull request to your repository from a fork, but no pipeline is triggered, then the user most likely is following a project fork on their personal account rather than the project itself of CircleCi, causing the jobs to trigger under the user’s personal account and not the organization account. To resolve this issue, have the user unfollow their fork of the project on CircleCI and instead follow the source project. This will trigger their jobs to run under the organization when they submit pull requests.
> 
> image
> 
> https://circleci.com/docs/2.0/oss/#example-open-source-projects

I have no idea how to unfollow my fork and how to follow this repo. I am "watching" both, I could "star" it too, but "follow" I do not see anywhere. 

Btw, I now see the other four CIs

swanderz:
@JCZuurmond if navigate to your "personal" GitHub repos by selecting "JCZuumond" from the top-left corner drop down, you should not see `dbt-sqlserver` as an active project. If should only show up when "dbt-msft" is selected? Fairly certain you don't have that anyway though. 🤷 

image

tox.ini
```diff
@@ -1,18 +1,8 @@
...
-[testenv:integration-azuresql]
-basepython = python3
-commands = /bin/bash -c '{envpython} -m pytest -v test/integration/azuresql.dbtspec'
-passenv = DBT_AZURESQL_DB DBT_AZURESQL_PWD DBT_AZURESQL_SERVER DBT_AZURESQL_UID
-deps =
-    -r{toxinidir}/requirements.txt
-    -e.
\ No newline at end of file
+  pytest~=6.2.2
+  pytest-dbt-adapter~=0.4.0
```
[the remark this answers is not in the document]
JCZuurmond: 1. Since we are only running it for one python version we could simply call pytest directly.
2. This is a tricky one. In general, I think there is benefit for consistency, it is easier for collaborating. However, I feel that it is "off" to create a copy of the test environment for each subset of test we would like to run, while you can use `{posargs}`. Resulting in that if you want to change the test environment in tox, you have to change it in many places (although a search an replace might solve that). I copied this mostly from the `tox` repo and docs.
3. See this [`tox.ini`](https://github.com/tox-dev/tox/blob/master/tox.ini), maybe for linting and for docs generation. But I think the "real" benefit is testing it against different python versions and for different OSs.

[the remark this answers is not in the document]
JCZuurmond: It makes the environments smaller and there is less to install. In the `dev_requirements` are also non-test related dependencies.

JCZuurmond:
@swanderz : I followed the `dbt-mst/dbt-sqlserver` in circleci. All jobs were showing at some point, but now I only see two again.

Furthermore, I have updated the PR to incorporate your feedback. Wo the other CIs I can not see if they are passing now. I have added a `dbt` environment to tox for running the dbt commands. 

@mikaelene  and @jtcohen6: what do you think about this to set-up?

JCZuurmond:
@swanderz : I am a bit confused. There is a step in `integration-sqlserver` which runs `which dbt` before it should be installed. Can you tell where it is coming from? It is the wrong dbt version, which we are not expecting to exists there.

JCZuurmond:
So, I tweaked the set-up some more. I think it makes sense to separate the tests into two stages:

1. code only checks/tests: unit tests, linting, black, etc. These are the tests we do given this repo only
2. integration tests: test how this package works with other tools. In this case it makes sense to prefer `python setup.py install` over `pip install -r requirements.txt`. If we test for the later case, we might install a dependency which is not in the `setup.py` and thus not available for the users who `pip install dbt-sqlserver`. `python setup.py install` would reflect better what users of this package encounter.

.circleci/config.yml
swanderz: @JCZuurmond this why the steps are currently failing. Somehow `dbt==0.18.1` is getting installed. perhaps the Circle CI python orb auto creates `virtualenvs`? Once we get this resolved this PR is good to go! Thanks again for all the work

```
#!/bin/bash -eo pipefail
cd test/integration
dbt compile --target sqlserver_local_userpass
```
```
Running with dbt=0.18.1
Encountered an error:
cannot import name 'AdapterResponse' from 'dbt.contracts.connection' (/usr/local/lib/python3.7/site-packages/dbt/contracts/connection.py)
Exited with code exit status 2
```

.circleci/config.yml
```diff
@@ -21,6 +21,9 @@ jobs:
           MSSQL_IP_ADDRESS: 0.0.0.0
     executor: python/default
     steps:
+      - run:
+          name: "run which dbt"
+          command: which dbt
```
swanderz, line 26: @JCZuurmond you're right, regarding your below comment. this should be removed.

> @swanderz : I am a bit confused. There is a step in integration-sqlserver which runs which dbt before it should be installed. Can you tell where it is coming from? It is the wrong dbt version, which we are not expecting to exists there.

```suggestion
```

.circleci/config.yml
```diff
@@ -78,25 +74,21 @@ jobs:
...
     docker: 
       - image: dataders/pyodbc:1.4
     executor: python/default 
     steps:
       - run: *install-git
       - checkout
-      - python/install-packages:
-          pkg-manager: pip
       - run:
           name: install gnupg
```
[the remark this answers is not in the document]
JCZuurmond: this is installed inside the tox environment or with `python setup.py install`

.circleci/config.yml
```diff
@@ -63,45 +66,33 @@ jobs:
...
-      # https://github.com/dbt-msft/dbt-sqlserver/pull/71
-      # - run:
-      #     name: cnxn -- Azure SQL - AAD Password
-      #     command: |
-      #       cd test/integration
-      #       dbt compile --target azuresql_aad
   integration-azuresql:
     docker:
       - image: dataders/pyodbc:1.4
     executor: python/default
```
swanderz, line 73: last thought is this line still needed? Not sure what it does anymore. esp because we're running bash commands on a python-base docker image
```suggestion
    executor: python/default
```

swanderz (approved):
I think we're ready to merge.  much cleaner now IMHO. @mikaelene -- I approve!

.circleci/config.yml
```diff
@@ -19,40 +19,42 @@ jobs:
           ACCEPT_EULA: 'yes'
           MSSQL_SA_PASSWORD: 5atyaNadella
```
mikaelene, line 20: Maybe not related to this PR but we should change the hardcoded connection info to env vars.

mikaelene (approved):
I am not a tox expert. Looks good to me and the tests seems to work. Always good to clean things up :-). We don't have that much code to keep clean.

.circleci/config.yml
swanderz: good point. for me it isn't so important bc we're spinning up a brand new SQL Server each time we run the test then destroying it after

Issue #25, opened 2020-05-30
Check if index exists before attempting to create it


infused-kim:
Currently if you use `create_nonclustered_index()` on an incremental table, the refresh fails, because the index already exists.

I understand the current pattern is to use a pre-hook with `drop_all_indexes_on_table()`, but that leads to all indexes having to be re-built on every refresh and that the indexes can't be used for the refresh merge.

Is there anything that prevents us from checking if an index exists in `create_nonclustered_index()` and skipping the creation if it does?

I have implemented it like this and so far it seems to be working well...
```sql
{% macro ii_create_clustered_index(columns, unique=False) -%}
    {{ log("Creating clustered index...") }}
    {% set idx_name = this.table + '__clustered_index_on_' + columns|join('_') %}

    if not exists(select * from sys.indexes 
                  where 
                    name = '{{ idx_name }}' and 
                    object_id = OBJECT_ID('{{ this }}')
    )
    begin

        create
        {% if unique -%}
        unique
        {% endif %}
        clustered index
            {{ idx_name }}
            on {{ this }} ({{ '[' + columns|join("], [") + ']' }})
    end
{%- endmacro %}


{% macro ii_create_nonclustered_index(columns, includes=False) %}
    {{ log("Creating nonclustered index...") }}
    {% set idx_name = this.table + '__index_on_' + columns|join('_') %}

    if not exists(select * from sys.indexes 
                  where 
                    name = '{{ idx_name }}' and 
                    object_id = OBJECT_ID('{{ this }}')
    )
    begin
        create nonclustered index {{ idx_name }}
            on {{ this }} ({{ '[' + columns|join("], [") + ']' }})
            {% if includes -%}
                include ({{ '[' + includes|join("], [") + ']' }})
            {% endif %}
    end
{% endmacro %}
```

alittlesliceoftom:
Am I right that this hasn't been implemented in : 

https://github.com/dbt-msft/dbt-sqlserver/blob/master/dbt/include/sqlserver/macros/indexes.sql

I'm about to test it out, I can raise as a PR if works well for me and useful ?

swanderz:
closed by #116, but related to #163

PR #116, opened 2021-02-21
Make index macros not recreate indexes on incremental models

infused-kim:
If you create indexes on an incremental model like this:
```jinja
{{
    config(
        materialized='incremental',
        as_columnstore = false,
        unique_key='my_key',
	    post_hook = [
		    "{{ create_clustered_index(columns = ['my_key'], 
                                                                 unique=True) }}",
		    "{{ create_nonclustered_index(columns = ['id']) }}",
	    ]
    )
}}
```

You get this error on subsequent runs:
```shell
  ('42S11', "[42S11] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]The operation failed because an index or statistics with name ‘my_table__clustered_index_on_my’_col already exists on table ‘my-db.dbt_xyz.my_table’. (1913) (SQLExecDirectW)")
```

This can be avoided by dropping the indexes in a pre-hook, but that leads to the index being recreated from scratch for all rows in the table. On a large incremental model that's probably not ideal in most cases.

This PR adds a check to the macros to only create the index if it doesn't exist yet. This way SQL server only adds the newly merged values to the index and doesn't reprocess all rows.

I don't think this would create any breaking changes for existing users of the macros. If they are already dropping indexes, it would not affect them.

And if they are using indexes on non-incremental models, then it should also not make a difference.

I have been using these macros for several months in production now and haven't had any issues.

This addresses the issue https://github.com/dbt-msft/dbt-sqlserver/issues/25 I raised about this.


alittlesliceoftom:
@mikaelene I think this one should be approved, I've been using the macros (just from my local macros folder) and they work really nicely. 

Big thanks @infused-kim for raising

mikaelene:
Great! @infused-kim , can you update the changeling as well? Adding a row of which issue it solves etc...

visch:
Works for me

@infused-kim  

Here's an example changelog to add, I can rebase if you'd like?

**New Features**

- Adding Index Macros now check to see if the index already exists before attempting to add the index to your table

semcha:
@swanderz Hi! What does it take to release this PR?

Issue #117, opened 2021-02-22
snapshot with unique key longer than 30 chars fails on duplicate


dubravcik:
Hi, I have a snapshot with unique_key on hash and I am getting duplicate error on internal dbt_scd_id in merge phase.

I checked the dbt code and problem is, it converts the id_hash to varchar. SQL Server uses by default 30 (!?wt.f!?) chars if you state varchar without a length.

```
CONVERT(VARCHAR(32), HashBytes(''MD5'', 
        coalesce(CONVERT(>>>>varchar<<<<, [id_hash], 121), '''')  + ''|'' +     
        coalesce(CONVERT(>>>>varchar<<<<, CONVERT(DATETIME2, ''2021-02-22 19:33:59.020000''), 121), '''') 
    ), 2)
```    

I would probably change it to varchar(max) as it is changed to varchar(32) at the end anyway.

mikaelene:
Relates to #108. @dubravcik do you think #118 will solve it?

dubravcik:
I don’t think so, it still uses `varchar` which defaults to 30 chars. I suggest `varchar(max)` , I don’t think that it will have a significant performance impact, but I can test it on my data.

mikaelene:
Ahh. I’ll change to max. There was also an issue when I tried to cast it as a date before. I somehow missed that it uses the key to.

mikaelene:
closed by #118

Issue #108, opened 2021-02-15
Issue with snapshots when unique key is varbinary (0.19)


b-per:
Hi!

Just testing the move from dbt-sqlserver=0.18 to dbt-sqlserver=0.19 and some snapshots that are working in v0.18 are failing now. This is the case when the unique_key of the snapshot is already hashed and of the type `varbinary`.

Here is the error: 
```
('42000', '[42000] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]The style 121 is not supported for conversions from varbinary to varchar. (9809) (SQLExecDirectW)')
```

The workaround I found is to change `unique_key='my_sk'` to `unique_key='CONVERT(NVARCHAR(32),my_sk,2)'` in the snapshot config but I am wondering if the snapshot macro couldn't be updated instead.

mikaelene:
Hi, I can understand the error, but not why you didn't get an error before. If I remember correct, dbt is making a new key called SCD_ID or similar that is a hash of converting the key you supplied "my_sk" with some date columns etc. It can probably be so that this is not supported by SQL Server. But why did it work before? This is the macro and I don't think it has changed: https://github.com/dbt-msft/dbt-sqlserver/blob/020be4a3ffef68c1af9a827bda27f3d0e5628901/dbt/include/sqlserver/macros/materializations/snapshot/strategies.sql#L1-L6

camfrout:
I just checked the `strategies.sql` file that I am using before the upgrade to v0.19 and it still has the "old" `cast` instead of `convert`, hence why I didn't have the error until now. I just checked `pip list` and this is the macro shipped with dbt-sqlserver=0.18.0 from pypi.

```
{% macro sqlserver__snapshot_hash_arguments(args) %}
    CONVERT(VARCHAR(32), HashBytes('MD5', {% for arg in args %}
        coalesce(cast({{ arg }} as varchar ), '') {% if not loop.last %} + '|' + {% endif %}
    {% endfor %}), 2)
{% endmacro %}
```

mikaelene:
Sorry! You are right! That was changed to increase the precision in the timestamp used in the snapshot. The previous code truncated the timestamp to second. This made the CI-tests fail for snapshots. Unlikely to happen in real life. https://github.com/dbt-msft/dbt-sqlserver/issues/61. I'll mark this as a bug. If you have an idea of how to solve this, please let me know.

camfrout:
[...]

Issue #61, opened 2020-11-20
dbt-adapter-test test_dbt_snapshot_strategy_check_cols failing


swanderz:
This is the only test (besides `data_test_ephemeral_models`) that isn't passing CI. [relevant Circle CI run](https://app.circleci.com/pipelines/github/dbt-msft/dbt-synapse/90/workflows/be73e2d3-f70f-45ef-976d-a1be93888a8f/jobs/97/steps) for those who belong to dbt-MSFT org. Here's [the full gist](https://gist.github.com/swanderz/3dc131dbd9b28e71e345e4354231fd8f) from CircleCI that includes all the SQL executed when only this test is run and no others. Below is the pytest result summary.

Perhaps the merge functionality isn't working properly and it's only inserting or only updating?

```python
============================= test session starts ==============================
platform linux -- Python 3.7.9, pytest-6.1.2, py-1.9.0, pluggy-0.13.1 -- /root/project/.tox/integration-synapse/bin/python
cachedir: .pytest_cache
rootdir: /root/project
plugins: dbt-adapter-0.3.0
collecting ... collected 1 item                                                               

test/integration/sqlserver.dbtspec::test_dbt_snapshot_strategy_check_cols FAILED [100%]

=================================== FAILURES ===================================
________________ usecase: test_dbt_snapshot_strategy_check_cols ________________
Unknown error: expected cc_all_snapshot to have 30 rows, but it has 20
assert Decimal('20') == 30
  +Decimal('20')
  -30 in test index 12 (item_type=relation_rows)
```

mikaelene:
The bug here was that we were using a dbt_scd_id of id + cast(get date() as varchar). The casting of get date removed the milliseconds. When the test runs the get date cast gets the same value as the previously inserted record so the merge that are looking for different dbt_scd_ids did not find it. The change above increases precision in the get_date cast and solvs the problem.

PR #118, opened 2021-02-22
Fix for bugs in snapshots

mikaelene:
When I (thought I) solved #61. I introduced a bug #108, #117. This should solve that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+### v0.19.0.2
+
+#### fixes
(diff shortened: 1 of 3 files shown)


swanderz (approved):
We've been avoiding using snapshots as best we can, so I can't try running this on existing snapshot tables. Might be worth cutting a beta release for users to try? or merge this to master and ask users to try snapshots after installing from github?

I am interested to know if this will have an impact on tsql-utils & dbt-utils. should [`tsql_utils.sqlserver__current_timestamp()`](https://github.com/dbt-msft/tsql-utils/blob/60fbeb3d8d4ef508fd703e34daa408243925d9bd/macros/dbt_utils/cross_db_utils/current_timestamp.sql#L1-L3) also be changed? I don't think it makes a big difference, but here's a [dbt-utils issue](https://github.com/fishtown-analytics/dbt-utils/issues/339) I created to hopefully learn more.

Issue #107, opened 2021-02-15
Remove pyodbc docker file


JCZuurmond:
Currently there is a docker file which installs the ODBC driver for Microsoft SQL server. The container created from the docker file is used in the Circle CI, only. This container/file is not needed if the ODBC driver is installed in a CircleCI (build) job.

- this docker file/container is redundant since we have circleci
- changing the docker container is cumbersome, since the publishing is not automated.
   The docker file also install the requirements.txt which is causing version problems and the requirements.txt does not exists if PR #105 is merged.

JCZuurmond:
I think it would make sense to make a `build` stage like [here](https://circleci.com/docs/2.0/sample-config/#sample-configuration-with-sequential-workflow)

PR #120, opened 2021-02-26
always build full env during CI

swanderz:
fixes: #107


.circleci/config.yml
```diff
@@ -5,48 +5,100 @@ orbs:
...
+            python3 -m venv venv
+            . venv/bin/activate
+            pip install -r dev_requirements.txt
+      # - python/install-packages:
+      #     pip-dependency-file: dev_requirements.txt
+      #     pkg-manager: pip
+      - persist_to_workspace:
+          root: .
+          paths:
+            - .
```
swanderz, line 55: @JCZuurmond, something tells me that the current dir isn't a broad enough scope to capture the debian package installs. I suspect if we can broaden the paths to be be a higher directory (or the whole container?!) then the below steps will work...

.circleci/config.yml
```diff
@@ -5,48 +5,100 @@ orbs:
...
-      - checkout
-      - run: tox -- -v test/unit
+      - attach_workspace:
+          at: .
+      - run:
+          name: unit test
+          command: |
+            sudo apt-get install unixodbc-dev
+            . venv/bin/activate
+            tox -- -v test/unit
```
swanderz, line 67: here's an example where I don't think the workspace is correctly persisted bc I have to install `unixodbc-dev` again to get things to work...

.circleci/config.yml
```diff
@@ -5,48 +5,100 @@ orbs:
...
 jobs:
+  build:
+    docker: &python3-7
+      - image: circleci/python:3.7.9
+    steps:
+      - checkout
+      - run:
+          name: install pyodbc deps
+          command: |
+            sudo apt-get update && \
```
JCZuurmond, line 16: I think the `&& \` or not needed inside the CI. 

This is a common practice to do with Docker files. They are build up from layers, one for each `RUN` command. You can reduce the number of layers by chaining the bash statements with `&& \`. You can also reduce the size of a layer if you finish with a clean up command. In this case there is not a benefit of doing this.

.circleci/config.yml
```diff
@@ -5,48 +5,100 @@ orbs:
...
+          name: TEMP -- downgrade security
+          command: |
+            sudo chmod +rwx /etc/ssl/openssl.cnf && \
+            # change TLS back to version 1
+            sudo sed -i 's/TLSv1.2/TLSv1/g' /etc/ssl/openssl.cnf && \
+            # allow weak certificates (certificate signed with SHA1)
+            # by downgrading OpenSSL security level from 2 to 1
+            sudo sed -i 's/SECLEVEL=2/SECLEVEL=1/g' /etc/ssl/openssl.cnf 
+      - run:
+          name: install python deps
```
JCZuurmond, line 44: We can use `tox` for this, so not needed really. Maybe `pip install dbt` to run the dbt commands

.circleci/config.yml
```diff
@@ -5,48 +5,100 @@ orbs:
...
     steps:
-      - checkout
-      - run: tox -- -v test/unit
+      - attach_workspace:
+          at: .
+      - run:
+          name: unit test
+          command: |
+            sudo apt-get install unixodbc-dev
+            . venv/bin/activate
```
JCZuurmond, line 66: `tox` is in itself a virtualenv so not needed.

.circleci/config.yml
JCZuurmond: It would be nice indeed to use the `unixodbc-dev` from the build. Let's investigate.

.circleci/config.yml
JCZuurmond: yes, we should locate where the packages are installed. Maybe the following path would already help

```
- "/usr/local/bin"
```

.circleci/config.yml
```diff
@@ -5,48 +5,100 @@ orbs:
...
+          name: install MSFT ODBC Driver
+          command: |
+            curl https://packages.microsoft.com/keys/microsoft.asc | sudo apt-key add - && \
+            curl https://packages.microsoft.com/config/debian/9/prod.list | sudo tee -a /etc/apt/sources.list.d/mssql-release.list && \
+            sudo apt-get update && \
+            sudo ACCEPT_EULA=Y apt-get install -y msodbcsql17 odbc-postgresql mssql-tools
+      - run:
+          name: config for MSFT driver 
+          command: |
+            export PATH="$PATH:/opt/mssql-tools/bin"
```
JCZuurmond, line 31: We could add this to the `BASH_ENV` so that it does not need to be reference later.

```
             echo 'export PATH="$PATH:/opt/mssql-tools/bin"' >> $BASH_ENV
```

.circleci/config.yml
```diff
@@ -5,48 +5,100 @@ orbs:
...
+  build:
+    docker: &python3-7
+      - image: circleci/python:3.7.9
+    steps:
+      - checkout
+      - run:
+          name: install pyodbc deps
+          command: |
+            sudo apt-get update && \
+            sudo apt-get install -y curl build-essential g++ apt-transport-https && \
```
JCZuurmond, line 17: it would be nice to see if `sudo` is not needed (for all installs)

.circleci/config.yml
JCZuurmond: Running the following for the packages name might help to locate where the installs are done:

```
dpkg -L <packagename>
```

JCZuurmond: and another good try is the `etc` folder:

```
- "/etc"
```

.circleci/config.yml
```diff
@@ -5,48 +5,104 @@ orbs:
...
+            python3 -m venv venv
+            . venv/bin/activate
+            pip install -r dev_requirements.txt
+      # - python/install-packages:
+      #     pip-dependency-file: dev_requirements.txt
+      #     pkg-manager: pip
+      - persist_to_workspace:
+          root: /.
+          paths:
+            - usr/bin/iusql
```
JCZuurmond, line 55: I think the leading `/` is missing here:  ` /usr/bin/isql `

swanderz: when i tried that, the error was, "please list paths relative to the "root" directory, i.e. `/.`

.circleci/config.yml
swanderz: I also would like to, but it seems CircleCi's python docker image doesn't give root user access for some reason

.circleci/config.yml
swanderz: we need a virtualenv so that `tox` itself is available as a dependency in all the downstream steps

swanderz:
@JCZuurmond thanks for all the feedback. The current error is about how downstream jobs are having trouble unzipping all the dirs, we're passing in the `persist_to_workspace` step. [The docs here](https://circleci.com/docs/2.0/configuration-reference/#persist_to_workspace) make me think you can't persist any files or dirs that already exist in the baseline executor Docker image.

> The workspace is additive only, jobs may add files to the workspace but cannot delete files from the workspace. Each job can only see content added to the workspace by the jobs that are upstream of it. When attaching a workspace the “layer” from each upstream job is applied in the order the upstream jobs appear in the workflow graph.

Beyond that, I'm skeptical, that it's even worthwhile, because all `persist_to_workspace` does is zip the given paths into a tarball, then unzip it into the same location for every downstream job. In our case, the paths together result in a 490 MB tarball which takes quite a bit of time to zip and unzip.

My new thinking is to take the linux driver install steps and roll that into a new `dbt-msft/msodbc_py` Docker image that's built on top of `circleci/python:3.7.9` (which is what Fishtown is IIRC). We still keep a build step, but it would only create a virtualenv that contains only what's needed to run:
- `tox`(for integration and unit testing ) and
- `dbt-sqlserver` for the connection tests (at least until we port those to `tox` as well).

JCZuurmond:
@swanderz : sounds like a good plan!

I think we should use [this base image](https://hub.docker.com/r/cimg/base) instead of the `circleci/python:3.7.9` one - since python is not needed in the build and we easily add python by specifying a secondary image. 

CircleCI has a [post](https://circleci.com/blog/announcing-our-next-generation-convenience-images-smaller-faster-more-deterministic/) on the "next-gen" images, which suggests the `circleci/python:3.7.9` is not the preferred choice anymore - the ones with `cimg/` are.

Also, a set-up like [this demo repo](https://github.com/CircleCI-Public/circleci-demo-docker) is nice. They have the `Dockerfile` in `.circleci/images/primary/` to make clear its purpose is to be used with CircleCI. They reference the image [here](https://github.com/CircleCI-Public/circleci-demo-docker/blob/f080cac0f844910b6416f8cbb3e40443c62e0251/.circleci/config.yml#L5). 

In theory we could also automate the [docker build](https://circleci.com/docs/2.0/building-docker-images/) in a circleci job, but I am not settled on if CircleCI recommends this.

JCZuurmond:
Furthermore, CircleCI has something called [orbs](https://circleci.com/orbs/). They have a [python orb](https://circleci.com/developer/orbs/orb/circleci/python), which might simplify the syntax for `pip install` and running `tox`.

When orbs are preferred over (primary) Docker images is a bit [unclear to me](https://discuss.circleci.com/t/orbs-vs-custom-images/27681). But I like that you can parameterize orbs more easily, for example, the following step would work for us:

``` yaml
      - python/install-packages:
          args: dbt
          pkg-manager: pip
```

swanderz:
closing in favor of #122

PR #122, opened 2021-03-02
Clean up CI config and Docker image

### LINKED ISSUE #107: Remove pyodbc docker file

swanderz:
- CircleCI config is now:
    - greatly cleaned up, and
    - installed the package and deps inline and no longer conflicts with version in Docker container.
- Docker image for CI now:
    - no longer installs the `dev_requirements.txt` or `dbt-sqlserver`
    - uses `cimg/python:3.7.10` instead of `docker/python:3.7-slim` as an upstream image.
    - is greatly simplfied due to reliance on CircleCI's official image

fixes #107


swanderz:
@JCZuurmond I'm almost there! Just having issues installing `dbt-sqlserver` outside of the virtual environment. I think I need to explicitly install and upgrade pip inside the docker image...

.circleci/config.yml
```diff
@@ -5,30 +5,41 @@ orbs:
   azure-cli: circleci/azure-cli@1.1.0
 
 jobs:
-  unit:
-    docker:
-      - image: dataders/pyodbc:1.4
+  build:
```
JCZuurmond, line 8: Given that we only pip install some packages, the build job is maybe not needed anymore. What could be nice, is using the python orb like [here](https://github.com/CircleCI-Public/python-orb/issues/70#issuecomment-788144946).

add to top level:

```
  orbs:
    python: circleci/python@1.3
```

And use the [python/install-packages](https://circleci.com/developer/orbs/orb/circleci/python#commands-install-packages) command to install the packages, for example:

```
        - python/install-packages:
            pkg-manager: pip
            args: tox
            pypi-cache: true
            venv-cache: false
            pre-install-steps:
              - run: pyenv install 3.7.8
              - run: pyenv install 2.7.18
              - run: pyenv local 3.8.7 3.7.8 2.7.18
```

This could be the first step of each job (change the installs when only `dbt` is needed) and with caching it would probably be fast. However, a build job would also be ok, but I would try to avoid the virtual env.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py37
+envlist = py36
```
JCZuurmond, line 2: why change this to python3.6?

JCZuurmond (commented):
I think the struggles with installing `dbt-sqlserver` occur because python is installed in the docker image. I think moving the python part out of the docker file should solve this

swanderz (commented):
@JCZuurmond ok now things are looking much cleaner IMHO. what do you think?

.circleci/config.yml
swanderz: I added back the `python/install-packages` command. but it can't work to replace `pip install .` AFAIK. regardless, could be a future PR if need be.

JCZuurmond:
yes, looks good @swanderz! I think it is ready to be merged.

There is one change I suggested before, but do not mind too much if not implemented, which I think is nice: use the `cimg/base` image in the Dockerfile and add `cimg/python` as secondary image in the `.circleci/config.yml`. This makes it more clear in the CI which python version is expected.

JCZuurmond (approved):

mikaelene (approved):
Looks great!

PR #124, opened 2021-03-30
Support for changing databases

### LINKED ISSUE #110: Changing the database via {{ config(database='OTHER_DB') }} doesn't work
My default database name is `EDW_BDM` but I'm trying to create this one view in a different database `FI_DM_EBI`.

sqlserver.sql:
```
{{ config(database='FI_DM_EBI') }}
SELECT
  'a' AS col1,
  'b' AS col2
```

`dbt run` prints `START view model FI_DM_EBI.dbo.sqlserver` and dbt docs show below database structure:

image

However, the view still gets created in `EDW_BDM`. Looking at debug output, the connection string used is the default one that has the database `EDW_BDM`.

The same thing happens when changing the database name in `dbt_project.yml`:
```
models:
  hopedw:
    example:
      database: FI_DM_EBI
```

Please add support for changing the database.

semcha:
Fixes #110.


semcha:
@swanderz @mikaelene Hi! How I can test this PR in azuresql?

swanderz:
@semcha I'm not sure! CircleCI should be testing this automatically...

swanderz:
@semcha you should re-open this! [This link](https://app.circleci.com/pipelines/github/dbt-msft/dbt-sqlserver/286/workflows/583baba4-49f8-45e7-8a14-e70edef1c712/jobs/1007) will show you how the integration tests are failing on Azure SQL, you just have to sign into Circle Ci with your GItHub account.

It's also quite easy to make a basic Azure SQL database on a free Azure subscription and use that to run the integration tests against it...

## issue

### error message
```
('42000', "[42000] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]
    Incorrect syntax near '-'. (102)
(SQLExecDirectW)")
```

### SQL query responsible for message

```sql
select
        table_catalog as [database],
        table_name as [name],
        table_schema as [schema],
        case
            when table_type = 'BASE TABLE' then 'table'
            else table_type
        end as [type]
    from **********************.information_schema.tables
    where table_schema = 'dbt_test_azure_sql_210330220928759444776398'
```

### root cause

looks like the bug right now might have to do with our current quoting policy because our database (starrred out) right now for security reasons, has a few `-`'s in it. The from line should probably be double-quoted like this:

```sql
from "**********************"."information_schema"."tables"
```

swanderz (commented):
@semcha you're closer than anyone's ever gotten before! if you accept my changes about properly quoting, the new error message is

> Reference to database and/or server name in `[****************].sys.schemas` is not supported in this version of SQL Server.

So here's two potential approaches:
1. try a `USE [{{relation.database}}]` at the top of these macros?
2. implement some tricky conditional logic based on the result of the `SERVERPROPERTY()` function. [here's a previous discussion](https://github.com/dbt-msft/dbt-synapse/issues/18#issuecomment-733920404).

@mikaelene what do you think? we're close I feel!

PR #111, opened 2021-02-17
Adding USE [{{relation.database}}] where appropriate in adapters.sql

panasenco:
Fixes #110. I messed with `adapters.sql` until changing databases became possible. Did a little bit of refactoring while I was at it...

I couldn't figure out how to run the unit tests so hoping the CI will reject the PR if something's wrong with it...

PR #126, opened 2021-03-31
Support for changing database in model configuration

### LINKED ISSUE #110: Changing the database via {{ config(database='OTHER_DB') }} doesn't work

semcha:
Fixes #110.


semcha:
@swanderz Why circleci run only unit test?

swanderz:
> @swanderz Why circleci run only unit test?

not sure it was saying "unauthorized". I was able to manually re-run the checks and they are passing now. but wanna take a look at this list and see if some might apply to you? You have a CircleCI account right?
https://support.circleci.com/hc/en-us/articles/360050273651-Builds-Unauthorized-due-to-contexts

dbt/include/sqlserver/macros/adapters.sql
```diff
@@ -24,17 +24,17 @@
       case when table_type = 'BASE TABLE' then 'table'
            when table_type = 'VIEW' then 'view'
            else table_type
-      end as table_type
+      end as [type]
```
swanderz, line 27: any reason you changed this?

```suggestion
      end as table_type
```

dbt/include/sqlserver/macros/adapters.sql
```diff
@@ -107,14 +109,27 @@
...
+    use [{{ relation.database }}];
+    {{ sqlserver__create_view_exec(relation, sql) }}
 {% endmacro %}
 
 
 {% macro sqlserver__rename_relation(from_relation, to_relation) -%}
   {% call statement('rename_relation') -%}
+    {% if from_relation.database != to_relation.database %}
+    {{ exceptions.raise_compiler_error("Can't do cross-database rename {} -> {}".format(from_relation, to_relation)) }}
+    {% endif %}
```
swanderz, line 131: @semcha was this prompted by an error?  I'm not sure why dbt would ever rename tables across databases? I'm pretty sure this macro is only used by dbt for renaming temp tables to their final names

dbt/include/sqlserver/macros/adapters.sql
```diff
@@ -107,14 +109,27 @@
...
+    ');
+{% endmacro %}
+
+
 {% macro sqlserver__create_view_as(relation, sql) -%}
-  create view {{ relation.schema }}.{{ relation.identifier }} as
-    {{ sql }}
+    use [{{ relation.database }}];
+    {{ sqlserver__create_view_exec(relation, sql) }}
 {% endmacro %}
```
swanderz, line 124: @semcha was the reason for this to make the macro run in a single transaction?

swanderz (commented):
The CircleCI tests are passing for Azure SQL, but I'm think I'm going to test this out over the next few days.

dbt/include/sqlserver/macros/adapters.sql
semcha: PostgreSQL adapter uses field name "type":
https://github.com/fishtown-analytics/dbt/blob/develop/plugins/postgres/dbt/include/postgres/macros/adapters.sql#L63

dbt/include/sqlserver/macros/adapters.sql
semcha: I took this code from https://github.com/dbt-msft/dbt-sqlserver/pull/111 , but I think it's not necessary if this macros only for dbt use only

dbt/include/sqlserver/macros/adapters.sql
semcha: What is your suggestion? :)

swanderz: I think my suggestion would be:
```sql
{% macro sqlserver__create_view_as(relation, sql) -%}
  use [{{ relation.database }}];
  create view {{ relation.include(database=false) }} as
    {{ sql }}
{% endmacro %}
```

dbt/include/sqlserver/macros/adapters.sql
swanderz: @mikaelene can you see this being valuable? My thinking is that this macro is only used for changing a table called `example__dbt_tmp` to `example` at the end of the materialization.

dbt/include/sqlserver/macros/adapters.sql
semcha: @swanderz That way you get an error `'CREATE VIEW' must be the first statement in a query batch.`

swanderz (commented):
@semcha have you tried running a dbt project after installing the adapter from this branch? I'd be interested if it works as we hope.

semcha:
@swanderz Yep, it's currently running on my production server

swanderz:
ok cool. so you're enabled in the short-term then. It's on my plate to see if this plays nicely with dbt-synapse. I hope to have time this week to do so. Thanks for all your work on this so far @semcha! this is the number one requested feature

mikaelene (approved):
Hi, This is all ok with me! @swanderz , do you have questions if this will work with Synapse to handle before merging?

swanderz:
> Hi, This is all ok with me! @swanderz , do you have questions if this will work with Synapse to handle before merging?

yeah this is the last piece. I should have time to check in the next few days

panasenco:
Thank you @semcha for this PR! I totally dropped the ball on this. I just experienced a `No item can be found` error that was fixed by switching to your branch, so your code is more reliable than mine. Great job, hope this gets merged soon!

panasenco:
Hi, could someone go ahead and merge this? :)

swanderz:
Hey @panasenco we definitely want to merge this, but I need to test these changes with dbt-synapse which I haven't had the time to do yet. I promise to do it this week sometime.

In the meantime, you can use this branch by installing this branch from GitHub or by cloning it locally and calling `pip install -e .` from the repo directory after checking out this branch.

panasenco:
Yeah, I've been doing that, so no worries. :)

dbt/include/sqlserver/macros/adapters.sql
```diff
@@ -43,7 +43,7 @@
 
 {% macro sqlserver__create_schema(relation) -%}
   {% call statement('create_schema') -%}
-    USE [{{ relation.database }}]
+    use [{{ relation.database }}];
```
swanderz, line 46:
```suggestion
    USE [{{ relation.database }}];
```

dbt/include/sqlserver/macros/adapters.sql
```diff
@@ -93,6 +84,7 @@
    {% set object_id_type = 'U' %}
    {%- else -%} invalid target name
    {% endif %}
+  use [{{ relation.database }}];
```
swanderz, line 87:
```suggestion
  USE [{{ relation.database }}];
```

dbt/include/sqlserver/macros/adapters.sql
```diff
@@ -107,14 +99,24 @@
...
+    execute('create view {{ relation.include(database=False) }} as
+    {{ temp_view_sql }}
+    ');
+{% endmacro %}
+
+
 {% macro sqlserver__create_view_as(relation, sql) -%}
-  create view {{ relation.schema }}.{{ relation.identifier }} as
-    {{ sql }}
+    use [{{ relation.database }}];
```
swanderz, line 112:
```suggestion
    USE [{{ relation.database }}];
```

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
 
+### v0.19.1
+
+#### features:
+
+- users can now delcare a model's database to be other than the one specified in the profile. This will only work for on-premise SQL Server and Azure SQL Managed Instance. [#126](https://github.com/dbt-msft/dbt-sqlserver/issues/126) thanks [@semcha](https://github.com/semcha)!
+
+#### under the hood
+
+- abandon four-part version names (`v0.19.0.2`) in favor of three-part version names because it isn't [SemVer](https://semver.org/) and it causes problems with the `~=` pip operator used dbt-synapse, a pacakge that depends on dbt-sqlserver
```
swanderz, line 11: @mikaelene this ok by you?

dbt/include/sqlserver/macros/adapters.sql
```diff
@@ -26,15 +26,15 @@
...
+    from [{{ schema_relation.database }}].information_schema.tables
     where table_schema like '{{ schema_relation.schema }}'
-      and table_catalog like '{{ schema_relation.database }}'
   {% endcall %}
   {{ return(load_result('list_relations_without_caching').table) }}
 {% endmacro %}
 
 {% macro sqlserver__list_schemas(database) %}
   {% call statement('list_schemas', fetch_result=True, auto_begin=False) -%}
+    use {{ database }};
```
swanderz, line 37:
```suggestion
    USE {{ database }};
```

dbt/include/sqlserver/macros/adapters.sql
```diff
@@ -107,14 +99,24 @@
...
-  create view {{ relation.schema }}.{{ relation.identifier }} as
-    {{ sql }}
+    USE [{{ relation.database }}];
+    {{ sqlserver__create_view_exec(relation, sql) }}
 {% endmacro %}
 
 
 {% macro sqlserver__rename_relation(from_relation, to_relation) -%}
   {% call statement('rename_relation') -%}
+    use [{{ to_relation.database }}];
```
swanderz, line 119:
```suggestion
    USE [{{ to_relation.database }}];
```

dbt/include/sqlserver/macros/adapters.sql
```diff
@@ -149,12 +152,13 @@
 
    {{ sqlserver__drop_relation_script(relation) }}
 
-   EXEC('create view {{ tmp_relation.schema }}.{{ tmp_relation.identifier }} as
+   use [{{ relation.database }}];
```
swanderz, line 155:
```suggestion
   USE [{{ relation.database }}];
```

swanderz (approved):
@semcha, I inititally made https://github.com/dbt-msft/dbt-synapse/pull/49 to prove to you that this PR, won't work only to realize that only a few changes would be necessary!

@mikaelene, this PR is ready to merge, provided you're okay with releasing a `v0.19.1` instead of a `v0.19.0.3`!

visch:
Just ran into this myself, wasted a bunch of time on it. I should have checked this PR first. Thanks for the work here!

swanderz:
I'll merge later today. @mikaelene will have to publish a release to PyPi though.

visch:
Smoke test.
My problem may be attached to #135 , but after updating the macros in my project manually I noticed some curious behavior.

One of my source tables isn't populated, the model looks like this

```
with source as (

        select
        *
        from {{ source('raw', 'nodata') }}    

), final AS (
	select 
	*
	from source
)
select * from final
```

DBT reports this as running successfully after updating the macro. Even though my raw data doesn't exist. Looking at the dbt log and running the generated sql manually the command fails with something like this

```
Msg 208, Level 16, State 1, Procedure sourcemodel__dbt_tmp_temp_view, Line 6 [Batch Start Line 0]
Invalid object name 'raw.nodata'.
Msg 208, Level 16, State 1, Line 33
Invalid object name 'sourcemodel__dbt_tmp_temp_view'.

Completion time: 2021-05-25T12:20:51.9277966-06:00
```

visch:
> Smoke test.
> My problem may be attached to #135 , but after updating the macros in my project manually I noticed some curious behavior.
> 
> One of my source tables isn't populated (I'd expect my staging model to fail) , the model looks like this
> 
> ```
> with source as (
> 
>         select
>         *
>         from {{ source('raw', 'nodata') }}    
> 
> ), final AS (
> 	select 
> 	*
> 	from source
> )
> select * from final
> ```
> 
> DBT reports this as running successfully after updating the macro. Even though my raw data doesn't exist. Looking at the dbt log and running the generated sql manually the command fails with something like this
> 
> ```
> Msg 208, Level 16, State 1, Procedure sourcemodel__dbt_tmp_temp_view, Line 6 [Batch Start Line 0]
> Invalid object name 'raw.nodata'.
> Msg 208, Level 16, State 1, Line 33
> Invalid object name 'sourcemodel__dbt_tmp_temp_view'.
> 
> Completion time: 2021-05-25T12:20:51.9277966-06:00
> ```

Just to be clear, this pull request makes the code better than it was before so I'd say go ahead! Everything works on my end, it's just an error caused by EXEC

SportsTribution:
I'm running into problems that are similar to visch's.
Arbitrary syntax errors in models, like writing SLECT instead of SELECT, go unnotized. dbt does not report an error at all — but no model is created. This problem does not exist in dbt-sqlserver=0.19.0.1 and must have been introduced later.
I'm writing this here, because I think the problem might be related to aforementioned code changes introducing multiple sql commands without a GO in between. 
https://github.com/flyway/flyway/issues/718 
> It seems that if you execute multiple sql commands in the same jdbc statement (which happens if you don't separate each sql command with GO), and an error occurs in some DML command (for example INSERT) that is not the first one in the batch, then the MS jdbc driver does not throw any exception if you don't call getMoreResults() on the jdbc statement, until the result of the failing command is retrieved.

To be clear, I think it's Microsoft's fault, not yours! ;)
 
All the best
Hannes

visch:
> I'm running into problems that are similar to visch's.
> Arbitrary syntax errors in models, like writing SLECT instead of SELECT, go unnotized. dbt does not report an error at all — but no model is created. This problem does not exist in dbt-sqlserver=0.19.0.1 and must have been introduced later.
> I'm writing this here, because I think the problem might be related to aforementioned code changes introducing multiple sql commands without a GO in between.
> [flyway/flyway#718](https://github.com/flyway/flyway/issues/718)
> 
> > It seems that if you execute multiple sql commands in the same jdbc statement (which happens if you don't separate each sql command with GO), and an error occurs in some DML command (for example INSERT) that is not the first one in the batch, then the MS jdbc driver does not throw any exception if you don't call getMoreResults() on the jdbc statement, until the result of the failing command is retrieved.
> 
> To be clear, I think it's Microsoft's fault, not yours! ;)
> 
> All the best
> Hannes

Yeah it's a problem. If you could put a test case together for them on https://github.com/dbt-msft/dbt-sqlserver/issues/135 I think it would help get to the bottom of it. I haven't had the time to prove myself :D

Issue #139, opened 2021-06-13
dbt doesn't properly return error message on wrong syntax


swanderz:
re-phrase of: #135 

using `dbt-sqlserver==0.19.1`, if you make a syntax error in a model, the dbt CLI does not return the error message like it used to in `dbt-sqlserver<0.19.1`.

This has me stumped so I'm tagging all the folks who know more about TSQL than me: @NandanHegde15 @mikaelene @semcha @panasenco

## steps to reproduce

1. configure a target of `type=sqlserver`
2. install dbt-sqlserver==0.19.1
3. create a model (`repro.sql`) with intentionally wrong sql
    ```sql
    -- models/repro.sql
    -- notice the 'x' after "SELECT"
    SELECTx 1 as col_name
    ```
4. run the new model `dbt run -m repro`
    ```
    (dbt) anders.swanson@AMAC02FG0TMMD6R SalesOps % dbt run -m repro  
    Running with dbt=0.19.1
    Found 21 models, 46 tests, 0 snapshots, 0 analyses, 712 macros, 0 operations, 1 seed file, 24 sources, 1 exposure

    16:07:15 | Concurrency: 1 threads (target='dev')
    16:07:15 | 
    16:07:15 | 1 of 1 START view model ajs.repro.................................... [RUN]
    16:07:16 | 1 of 1 OK created view model ajs.repro............................... [OK in 1.27s]
    16:07:16 | 
    16:07:16 | Finished running 1 view model in 8.44s.

    Completed successfully

    Done. PASS=1 WARN=0 ERROR=0 SKIP=0 TOTAL=1
    ```

## more info

in the full extract of the [dbt log](https://gist.github.com/swanderz/1f71ddb2eb45750c3903b06851847a44) you can see this query batch get executed
```sql
USE [dbt-msft-serverless-db];
execute(
    'create view "test"."repro__dbt_tmp" as
    SELECTx 1 as col_name'
);
```

Executing the above query batch directly against the database provides the following error message, that isn't returned by dbt.
```
Msg 102, Level 15, State 1, Procedure demo__dbt_tmp, Line 2
Incorrect syntax near 'SELECTx'.
```

Interestingly enough, looking at the db's logs, it _does_ show the error/failure.
```
<batch_information><failure_reason>Err 102, Level 15, Server dbt-msft-server
Incorrect syntax near 'SELECTx'.</failure_reason></batch_information>
```

## original motivation for change

[...]

Issue #135, opened 2021-05-19
Some SQL commands don't report a failure


visch:
EXEC('ALTER TABLE dbo.stg_things addd primary key(thing_id)')

This command will fail due to **addd**  being incorrect. This will just keep chugging along, is that just a side affect of EXEC that we have to live with? I'm going to attempt to not use EXEC, but I was warry when I read https://github.com/dbt-msft/dbt-sqlserver/issues/80 

Still new to DBT so I'm not certain how everything's supposed to work, but this almost certainly shouldn't happen. If a SQL command fails something should throw an error

visch:
Changed the title as I just had this SQL run and fail, but dbt reported everything to be ok

```
   ALTER TABLE tablename
         ADD
         SysStartTime DATETIME2 GENERATED ALWAYS AS ROW START HIDDEN
            CONSTRAINT DF_SysStart DEFAULT SYSUTCDATETIME()
         ,SysEndTime DATETIME2 GENERATED ALWAYS AS ROW END HIDDEN
            CONSTRAINT DF_SysEnd DEFAULT CONVERT(DATETIME2, '9999-12-31 23:59:59.9999999'),
         PERIOD FOR SYSTEM_TIME (SysStartTime, SysEndTime);
         
         ALTER TABLE tablename 
         SET (SYSTEM_VERSIONING = ON (HISTORY_TABLE = tablename_history));
         ```


The failure was due to another table already having the DF_SysStart constraint setup. This can be replicated by setting up a table with this constraint before running the query.

swanderz:
@visch very glad that you're putting dbt-sqlserver through it's paces here. my question is how are you running the code above? with a `post-hook`?

in [our related Slack thread](https://getdbt.slack.com/archives/CMRMDDQ9W/p1621019815068300), I mentioned how it might make sense to alter the create table macro to include an option to set `SYSTEM_VERSIONING`. Here's [the macro](https://github.com/dbt-msft/dbt-sqlserver/blob/83b8357519c585e4103481a5b987f0da6b376cce/dbt/include/sqlserver/macros/adapters.sql#L141-L165) you would need to change.

My problem is that dbt is all about using DML instead of DDL, which is what you'd need to create a Temporal Table...

visch:
Yeah no problem glad to use this, I appreciate all the work everyone's put into this to allow me to get sql working very easily. Just a few hiccups :D 

[...]

PR #140, opened 2021-06-15
catch error in multi-query batches

swanderz:
Fixes: #139, #135

#124 modified core adapter macros (e.g.`create_view_as()`) to be multi-query batches that were previously single queries.

This PR attempts to use @jeroen-mostert's solution to iteration through each query in a query batch looking for errors


swanderz:
@jtcohen6, I'm struggling with how to fetch then handle the pyodbc's `ProgrammingError` exception. More generally I'm trying to understand how the `SQLConnectionManager` class works with query batches...

## current approach to fixing
I've implemented @jeroen-mostert's approach. I'm not exactly sure what `cursor.nextset()` does, but when you call it on a batch query where the 2nd query has a syntax error, the pyodbc `Programming Error` exception pops up... This error doesn't surface if you only call `query.exectue()`

this PR now surfaces an "unhandled" error, but doesn't return the `Database Error` line about where the error occurred.

image

## desired (and previous state)
this is what the output _should_ look like using a version of dbt-sqlserver before the adapter macros had query batches.


image

## alternative solution split the queries and add/execute them independently?

One approach would be to implement something like [`SnowflakeConnectionManager._split_queries()`](https://github.com/fishtown-analytics/dbt/blob/14507a283e6acd710363aab9fc8fbcac1658002d/plugins/snowflake/dbt/adapters/snowflake/connections.py#L279-L285), but my reservations are:
1. would this work for an adapter macro like [`sqlserver__create_view_as()`](https://github.com/dbt-msft/dbt-sqlserver/blob/679b3515e9f056879c6609c78bb62791716c450a/dbt/include/sqlserver/macros/adapters.sql#L103-L114)
1. will executing a database switch command in two different batches work (I have to test this w/ a SQL Server instance)
2. where does `snowflake.connector.util_text.split_statements()` come from?

jeroen-mostert:
If it helps (actually I'm pretty sure it doesn't but I'm putting it out here anyway :-)) there's something weird with how pyodbc treats multi-statement batches for SQL Server. Calling `.nextset()` advances to the next result set of a batch -- this is a general mechanism surfaced also in, for example, .NET's `SqlDataReader.NextResult()`. However, this should not generally be necessary for statements that do not produce result sets -- `USE [database]; SELECT 1 a` is an example of a batch with two statements that produces *one* result set, not two. In .NET you would never need to call `.NextResult()` to make this batch return, but for some reason pyodbc will treat every statement result as a distinct result set, even if the statement just completed and nothing more. And instead of surfacing execution errors, it will treat an error that results from `.nextset()` as a programming error (?). Basically, pyodbc demands that you know *in advance* exactly how many statements you have and what each of them will return *even if they don't return anything*, which is quite a nasty restriction that I haven't seen in DB libs for other languages.

All this is not to suggest this is something that could or should be changed in the pyodbc layer just to make dbt happy, of course, since that might be quite impactful; executing individual statements is still the safest way to go. A `USE database` persists on the connection across batches, provided of course that 1) you do not close it and open a new one and 2) you do *not* use a separate `EXEC` for the statement, as that gets its own context and the DB is reverted at the end. (`EXEC 'USE db; <other stuff>'` works, of course, but you'll probably run into the same multi-statement result set gotchas again.)

Parsing batches and splitting them manually on anything but a dedicated batch separator like good old `GO` requires a T-SQL parser with full fidelity if there's any chance your statements contain escaped or quoted semicolons anywhere; I wouldn't recommend that path. T-SQL's a hideously byzantine language and the fact that the statement separator is mostly optional doesn't help; you really want to split your statements on a higher level if possible.

semcha:
Awesome fix! 👍 Faced with this today in my work with snapshot functionality.

jtcohen6:
@jeroen-mostert I really find myself agreeing. It's not good how pyodbc treats errors from multi-statement batches. It's also preferable, wherever possible, to supply individual queries one at a time, so long as they can be executed within a single transaction/session. `EXEC` perhaps can't be helped. Lacking other alternatives, pushing the error to stdout, even if imperfectly, is still a huge improvement.

> where does `snowflake.connector.util_text.split_statements()` come from?

That comes from the Snowflake python connector [here](https://github.com/snowflakedb/snowflake-connector-python/blob/0f7b9385adbd54e2e2608f3672528959ae943094/src/snowflake/connector/util_text.py#L32). The logic in this method is somewhere between "just split on semicolon" and "SQL parser with full fidelity." If I remember right, the Snowflake python connector does not properly handle multiple query statements at once (semicolon-delimited), so this is what we have to use instead.

dbt/adapters/sqlserver/connections.py
```diff
@@ -347,7 +347,7 @@ def add_query(self, sql, auto_begin=True, bindings=None, abridge_sql_log=False):
 
             # pyodbc does not handle a None type binding!
             if bindings is None:
-                cursor.execute(sql)
+                cursor2 = cursor.execute(sql)
```
mikaelene, line 350: Should it be called cursor2 or is this work in progress? Looked through the code and cursor2 will not be returned in the end.

visch:
Thank you for pushing this up! 

Sorry I didn't give better instructions for how to reproduce the error :D

So far (even with the cursor2 oddity) runs with this new setup, and I'm seeing failures where I would expect now. And I'm catching errors I was missing :O 


(installed on a new venv with 
`pip install git+https://github.com/dbt-msft/dbt-sqlserver.git@cb12056e18ad60859eeaa34217bc3a8762551d02`  , is there an easier way to install over an existing installation?```
 pip install --upgrade 
git+https://github.com/dbt-msft/dbt-sqlserver.git@cb12056e18ad60859eeaa34217bc3a8762551d02
``` wasn't doing it for me, and cloning and running `python setup.py install `wasn't either (it looked like pip thought the existing package was the most recent)

**Comment by visch** [comment]
Noticed another regression but it's not related to this (Caught because of this fix though).  https://github.com/dbt-msft/dbt-sqlserver/issues/143

File: dbt/adapters/sqlserver/connections.py (hunk 1 above)
Comment (swanderz) [reply dbt/adapters/sqlserver/connections.py:350]: good catch. I _think_ we don't need to return the object

```suggestion
                cursor2 = cursor.execute(sql)
```

visch:
Using this right now in my environment. So far so good, I've caught 20 plus errors I was missing :O

swanderz:
@jeroen-mostert, you are my hero -- not only for the fix but the depth of knowledge you were able to share. You certainly make me feel a lot saner, now that I know that both `pyodbc` and TSQL are at fault here.

That said, these EXEC statements are only needed to support the multi-database scenario that is only possible with on-premise SQL Server. There's a part of me that wants to kick this behavior into it's own adapter (`dbt-sqlserver-multi-db`) so single-database SQL Server users and Azure SQL users don't have to experience this workaround stdout behavior. @jtcohen6 @periclesrocha, the database version telemetry MSFT has on connections coming from dbt would greatly inform the right path to take...

For now, we should definitely merge and release this to everyone. @mikaelene would you do the honors of releasing a `v0.19.2` once this is merged?

PR #141, opened 2021-06-15
v0.20.0

swanderz:
- upgrade to `v0.20.0`
- also switched to recursive include


swanderz:
@jtcohen6 hate to bug you on this as I'm sure it's a tiny mistake I'm making here, but I'm struggling to figure out why `default__get_test_sql` is being used rather than `sqlserver__get_test_sql()`...

jtcohen6:
@swanderz The `get_test_sql` change (https://github.com/fishtown-analytics/dbt/pull/3461) will be going out in v0.20.0rc2 (which we hope to release end of this week, or very early next)

swanderz:
> @swanderz The `get_test_sql` change ([fishtown-analytics/dbt#3461](https://github.com/fishtown-analytics/dbt/pull/3461)) will be going out in v0.20.0rc2 (which we hope to release end of this week, or very early next)

🤦🏻 . Ok so that explains the CircleCI bug. but in my local testing I'm still seeing the issue and I'm installing dbt from GitHub on the `develop` branch...

dbt/include/sqlserver/macros/materializations/test.sql
```diff
@@ -0,0 +1,12 @@
+{% macro sqlserver__get_test_sql(main_sql, fail_calc, warn_if, error_if, limit) -%}
+    select
+      {{ fail_calc }} as failures,
+      case when {{ fail_calc }} {{ warn_if }}
+        then 1 else 0 end as should_warn,
+      case when {{ fail_calc }} {{ error_if }}
+        then 1 else 0 end as should_error
```
jtcohen6, line 7:
```suggestion
      case when {{ fail_calc }} {{ warn_if }}
        then true else false end as should_warn,
      case when {{ fail_calc }} {{ error_if }}
        then true else false end as should_error
```
dbt [expects these](https://github.com/fishtown-analytics/dbt/blob/c794600242e734d870d4bf1c8fb4c0349ab961eb/core/dbt/task/test.py#L36-L37) to be boolean. This may work by type coercion, I'm not sure

jtcohen6 (commented):
I was able to test this locally, implementing `postgres__get_test_sql` inside the `dbt-postgres` adapter, and it worked just fine. Same if put `postgres__get_test_sql` in my root project.

But it _doesn't_ work if I install dbt via `pip install git+https://github.com/fishtown-analytics/dbt.git`, even when I specify the `develop` branch. I'm not sure why...

dbt/include/sqlserver/macros/materializations/test.sql
swanderz: cool cool cool. but what if our favorite "byzantine" (h/t  @jeroen-mostert lol) dialect doesn't support a boolean type? all it has is `BIT`... Lemme try the local dev mode install and get back to you with what I find out

https://stackoverflow.com/a/32150794/3842610

jtcohen6: Ah... in that case, 0/1 may work, "true"/"false" (strings) may work, it just needs to be close enough to validate [here](https://github.com/fishtown-analytics/dbt/blob/c794600242e734d870d4bf1c8fb4c0349ab961eb/core/dbt/task/test.py#L108)

PR #147, opened 2021-06-24
fix for empty seed tables

### LINKED ISSUE #146: Seed file without data rows causes "max() arg is an empty sequence" error
Using a seed file containing only column headers and the dbt-snowflake adapter, the command "dbt seed --select seedfilename --full-refresh" creates an empty seedfilename table in Snowflake. With the dbt-sqlserver adapter v0.19.1, however, I get a "max() arg is an empty sequence" error and no table is created in MSSQL.

The error occurs in dbt/adapters/sqlserver/impl,py line 22. I am working around the error by replacing this line with
`max_len = max(lens) if len(list(column.values_without_nulls())) else 64`

swanderz:
fixes #146

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 ### v0.19.2
 
+#### fixes
+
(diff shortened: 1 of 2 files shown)


DVTimWilson (commented):
Hi Anders, the proposed fix resolves our issue, thank you. How soon might this be merged and released?

swanderz:
@DVTimWilson merged just now.`v0.19.2` should be released soon -- there's a number of other bugs our users are experiencing.
cc: @mikaelene

PR #150, opened 2021-07-05
Workarounds in create_table_as() macro for 'invalid target name' errors

DVTimWilson:
Hi folks, workarounds for issues I am experiencing when using materializations, both dbt's and custom, when working with new, not existing, regular and temporary tables. These cause 'invalid target name' errors when calling the 'sqlserver__drop_relation_script(relation)' macro. The workarounds have resolved all the issues with my use cases so far, but I appreciate they might not work for everyone so would warrant further investigation perhaps. Thank you.


DVTimWilson:
Hmmm, failing CircleCI Azure SQL test on USE statements that were already in the code. Sorry, but I know nothing at all about Azure SQL!

DVTimWilson:
I am going to close my pull request because it looks like I have further use case issues to investigate around the create_table_as() macro.

PR #175, opened 2021-11-08
fix index with columns containing spaces

dubravcik:
if columns contains space then index name built from the name contains space as well. Fixed with space replaced to underscore.


swanderz:
thanks @dubravcik!

PR #178, opened 2021-11-10
Add support for more authentication methods

sdebruyn:
I've been struggling with the current authentication methods and they always require a little bit of configuration effort while authentication with other `azure-identity`-based packages usually works quite easily.

So I've added 3 new ways to authenticate to your Azure databases:

* Automatic (really the easiest way)
* By reading environment variables
* Managed identity

All methods are properly documented on https://docs.microsoft.com/en-us/python/api/overview/azure/identity-readme?view=azure-python but I've updated the README as well.

I added tests for everything except managed identity since that is impossible to test without an actual managed identity configured.

I removed the storage of client ID & secret in the service principal authentication as environment variables to avoid side effects from using `DefaultAzureCredential` and to avoid leaking credentials outside of this context.


JCZuurmond (approved):
LGTM

I see you ran black, which is fine with me. Maybe @swanderz  we should add this in a pre-commit hook and test for it in the pipeline see #72

swanderz (changes requested):
@sdebruyn this looks great! happy to merge and release a new version with this change ASAP. can you please first:
- increment the version in `dbt/adapters/sqlserver/__version__.py` to `0.21.1`?
- add an entry for `v0.21.1` to `CHANGELOG.md`?

sdebruyn:
@swanderz done :)

swanderz (approved):
thanks @sdebruyn!!!

PR #190, opened 2021-12-01
fix secret in circleci config

sdebruyn:
Looks like I somehow made I typo in the name of the secret as it seems to be `DBT_AZURE_SP_SECRET` instead of `DBT_AZURE_SECRET`


swanderz:
thanks @sdebruyn!

PR #198, opened 2022-01-17
set env vars explicitly to fix circle ci check

sdebruyn:
I don't have any experience with Circle CI, but the point is that it should have the same env vars set as in the check above.


dataders:
thanks @sdebruyn!

PR #200, opened 2022-01-23
Implementing on_schema_change

semcha:
https://docs.getdbt.com/docs/building-a-dbt-project/building-models/configuring-incremental-models#what-if-the-columns-of-my-incremental-model-change


sdebruyn:
Should be rebased on latest master

Issue #195, opened 2021-12-23
Deadlocks on target server when more than 1 dbt thread


deanja:
# Expected Behaviour

dbt can run many tests in parallel without causing database deadlocks.
dbt can query information_schema and do DDL on separate threads without deadlocks.

# Actual Behaviour

Intermittent deadlocks.  So far only observed on `dbt build` (with tests) and `dbt test`.

```
('40001', '[40001] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Transaction (Process ID 62) was deadlocked on lock resources with another process and has been chosen as the deadlock victim. Rerun the transaction. (1205) (SQLFetch)')
```
An example deadlock log, showing implicated queries/ddl,  is pasted at the end of this issue.

# Steps to reproduce
```
dbt test --threads 8
```
or
```
dbt build--threads 8
```

Yesterday I could reliably get it to fail with 2 threads.  Today it's running fine with 8 sometimes.   Oooouch!    Seems slightly easier to recreate when there are some bigger relations (10M+ rows) to test.

## Configuration
A set of small generic dbt tests - I have ~50 tests 
DAG does not affect `dbt test`. For `dbt build` I have ~4 paths through the DAG.
dbt 1.0.0
SQL Server 2019
ODBC Driver 2017

# Possible causes

The "what changed recently " cause is that from dbt 0.21.1 `test` creates a lot more tables, as a workaround for there being no nested CTE functionality in SQL Server.

Cause in SQL Server.  Not sure, but seems the engine has it's own means of ensuring integrity when querying INFORMATION_SCHEMA and making DDL changes at the same time.  In DML, the query would simply get blocked (delayed) while the write completes, not cause a deadlock, except in some edge cases involving index updates.   Things were fine pre 0.21.1 probably because of lower volume of operations per second - most tests are quick, and they can take advantage of parallelism.  

A root cause is that there's not nested CTE in SQL Server.  dbt-core tests and packages like dbt-expectations increasingly use nested CTEs.

# Other information

Current workaround: limit dbt threads to 1.

Some discussion of possible causes and solutions on [dbt slack](https://app.slack.com/client/T0VLPD22H/CMRMDDQ9W/thread/CMRMDDQ9W-1640064902.122500)

[...]

PR #204, opened 2022-01-27
Add nolock

Elliot2718:
Adds with (nolock) to list_relations_without_caching macro. Fixes deadlock issues when running dbt tests in issue #195. Also see Slack thread here: https://getdbt.slack.com/archives/CMRMDDQ9W/p1640064902122500

# Outstanding questions:
Are we sure adding nolock is acceptable here? My understanding is that each dbt test is independent and does not use the same database objects, which should make this fine. There could be something I'm missing though.


deanja:
I am torn on this one.      I am a user of dbt and SQL Server, not a dbt core or adapter developer.

The PR fixes the issue.  No deadlocks when doing `dbt test` with 4 treads and ~150 tests, which all run in 7 seconds.   Without this fix it would cause multiple deadlocks.

Like you, I'm concerned about introducing `nolock` to the project. It's usually better to understand if/how the application (in this case dbt) is causing the deadlocks. and try to prevent them there.   It could be harder in this case because SQL Server might have its own quirks about concurrency of DDL operations and reads on information_schema.  

A agree that use of `nolock` in this particular macro (`sqlserver__list_relations_without_caching`) is probably safe to use in `dbt test`, because the tests are independent of each other.   It is probably also safe in  `dbt run`  and `dbt build` (**which also call that macro**) because they are constrained by dbt to run in DAG order.  They are unlikely to be sensitive to uncommitted rows/deletes from another transaction being returned by READ_UNCOMMITTED, which is the behaviour invoked by `nolock`.     

Are their other callers of that macro?     

Do we need to consider MS db platform compatibility? `nolock` is [supported ](https://docs.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-table?view=sql-server-ver15) on SQL Server and Azure SQL Database.  Does `dbt-sqlserver` target any other MS data platforms?

mikaelene:
Maybe this should be optional? Perhaps set in profiles.yml.

deanja:
Good  idea.
And/or only use nowait when threads > 1

Either way gives users/developers a heads up that 'nowait' is used as a
concession for multithreading.  As profile/cli doco, or simply the presence
of 'if' condition in the macro.

sdebruyn:
Needs rebase on latest master

SportsTribution:
Hi folks,

> Like you, I'm concerned about introducing nolock to the project. It's usually better to understand if/how the application (in this case dbt) is causing the deadlocks. and try to prevent them there. It could be harder in this case because SQL Server might have its own quirks about concurrency of DDL operations and reads on information_schema.

Not sure, I am of much help, but the locks seem to always be on 
`database_name.sys.sysschobjs`
and involves more or less
```
select
   table_catalog as [database],
   table_name as [name],
   table_schema as [schema],
   case when table_type = 'BASE TABLE' then 'table'
else table_type
   end as table_type
  from [database_name].information_schema.tables
 where table_schema like 'dbt_test__audit'
```
and things like
```
USE [database_name];   if object_id ('"dbt_test__audit"."unique_table_name1_key_temp_view"','V') is not null
   begin
   drop view "dbt_test__audit"."unique_table_name1_key_temp_view"
   end
```
... long story short, I also cannot come up with a better solution than nolock on the list_relations_without_caching - (even though I don't know why the upper statement can not simply what that the lower one is finished. But it's getting late).
Anyways, maybe this PR could proceed?

Cheers
Hannes

alieus:
@mikaelene @sdebruyn @dataders - any updates/suggestions on this PR?

mikaelene:
I've not been super involved in the development of this adapter lately. But when running tests, you usually don't update any data or definitions so for that reason I think it's perfectly fine to use "with no lock". I don't know how people use this adapter these days but my aim when creating it was to implement cloudDW-like behaviour on SQL server and the other clouds doesn't use lock what I know of. In an OLTP-setting it would be a bad idea, but as for DW and querying data for tests I guess it would be ok. Looks like the CI build is failing. If someone fix that I think we can merge.

sdebruyn:
I managed to fix some other deadlocks as well in #368 but I definitely think we need this one as well

PR #207, opened 2022-01-29
feat: adding support for bastion / jumpbox support

hugoJuhel:
# Objective
The PR aims at adding supports for Jump / Bastions servers to the connector so that connecting to a server behind a VPN would be possible.

# Proposed modifications
* The code modifies the way the connection is opened by the adapter.
* If provided with `ssh_tunnel_host`, `ssh_tunnel_port`, `ssh_tunnel_username` and one of either `ssh_tunnel_password` or `ssh_tunnel_ssh_key_path`, the adapter first establishes a `sshtunnel` to the jump host before connecting to the server.
* The code uses `sshtunnel` to handles the tunnel creation.
* A dataclass is introduced to bind the `pyodbc.Connection.close` method to the `SSHTunnelForwarder.close` method.
* The dataclass is preferred over registering a callback in `pyodbc` since `pyodbc.close` attribute is not settable.

### DIFF
--- a/dbt/adapters/sqlserver/connections.py
+++ b/dbt/adapters/sqlserver/connections.py
@@ -49,10 +49,11 @@ class SQLServerCredentials(Credentials):
     authentication: Optional[str] = "sql"
     encrypt: Optional[bool] = False
     trust_cert: Optional[bool] = False
-    jumpbox_ip: Optional[str] = None
-    jumpbox_port: Optional[int] = 22
-    jumpbox_username: Optional[str] = None
-    jumpbox_password: Optional[str] = None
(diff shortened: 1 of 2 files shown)


sdebruyn:
Should be rebased on current master so that there are no formatting changes anymore

hugoJuhel:
@sdebruyn 
Hi,

I have just pushed a commits with everything rebased on master.

hj

sdebruyn:
I'm not sure support for SSH tunneling should be a feature of this adapter that we would want to maintain, as well as an extra dependency that would not be needed for 99% of the users. Closing this for now.

PR #209, opened 2022-02-18
Add possibility to define query hints for models

l-j:
**Background:**
In my organization sometimes is useful to specify [T-SQL query hints](https://docs.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-query?view=sql-server-ver15) to improve query performance. Unfortunately, as far as I know with the current design of model materialization it is not possible to simply specify hints as part of our models, because later they are first created as views, and we cannot pass anything to `SELECT * INTO ... FROM ...` data loading part.

**New feature:**
In this PR I would like to add a new parameter `query_hints` to model config, which gives the ability to specify query hints according to needs when models are materialized as `table` and `incremental`.

**Usage examples:**
```
{{ config(query_hints="MAXDOP 1") }}
```
Multiple query hints can be provided using commas:
```
{{ config(query_hints="MAXDOP 1, USE HINT('QUERY_OPTIMIZER_COMPATIBILITY_LEVEL_110')") }}
```
I have tested this feature with Azure SQL Database.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,14 +192,21 @@ vars:
   max_batch_size: 200 # Any integer less than  or equal to 2100 will do.
 ```
 
-### Type hints
+### Materialization Query Hints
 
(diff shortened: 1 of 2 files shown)


sdebruyn:
@l-j Shouldn't we add the same functionality to views as well?

PR #215, opened 2022-04-17
Escape schema names so they can contain strange characters

johnf:


sdebruyn:
@johnf Thank you for your contribution! :)

sdebruyn (approved):

PR #216, opened 2022-04-20
switch to new adapter test framework

dataders:
* Uses new test adapter framework created by dbt Labs
* Uses tox for all tests
* Moves tests from Circle CI to GitHub actions
* Fixes issue with duplicate column names in tempdb


dbt/include/sqlserver/macros/adapters/relation.sql
```diff
@@ -7,6 +7,7 @@
 {% endmacro %}
 
 {% macro sqlserver__drop_relation(relation) -%}
+  {{ log('yyy relation: ' ~ relation, info=True) }}
```
ChenyuLInx, line 10: There's still log here

sdebruyn:
We should probably look into the thing where a table appears multiple times in tempdb but from a different schema.

dbt/include/sqlserver/macros/adapters/columns.sql
```diff
@@ -17,7 +17,7 @@
           from [{{ relation.database }}].INFORMATION_SCHEMA.COLUMNS
           where table_name = '{{ relation.identifier }}'
             and table_schema = '{{ relation.schema }}'
-          UNION ALL
```
dataders, line 20: OMG WOW great find!

PR #228, opened 2022-05-21
Pre-commit setup

sdebruyn:
Setting up pre-commit for this repo. It works well except for mypy which complains that `dbt-sqlserver` is not a valid package name.


dataders:
@sdebruyn do docs on how the pre-commit works? i'm a pre-commit newb. perhaps we make additions to the `CONTRIBUTING.md`?

PR #238, opened 2022-06-02
Add get_true_sql function for sqlserver adapter to allow snapshots to…

jakemcaferty:
Snapshots currently break when adding new columns to SQL Server.  A recent update to dbt-core changes the snapshot logic to use an adapter for new columns so that different languages can correctly handle a where clause that is always true.


jakemcaferty:
The specific dbt-core PR that this update will work with is below:
https://github.com/dbt-labs/dbt-core/pull/4871/files

sdebruyn:
@jakemcaferty Thank you for your contribution!

PR #239, opened 2022-06-02
run CI tests on trusted PRs and include secrets

sdebruyn:
CI tests were failing on PRs from external contributors because the secrets were not available

this changes that so that the tests without secrets always run and the tests needing secrets are only triggered when you add the label `safe to test`

### DIFF
--- a/.github/workflows/integration-tests-azure.yml
+++ b/.github/workflows/integration-tests-azure.yml
@@ -1,43 +1,13 @@
 ---
-name: Integration tests
+name: Integration tests on Azure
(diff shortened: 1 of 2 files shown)


dataders (approved):
sweet! just wondering -- what tests are not sensitive tests? my gut tells me the "local" sqlserver tests need not be sensitive

sdebruyn:
@dataders All tests requiring access to our Azure env should be approved by a maintainer to avoid a pull request becoming a pwn request. The other ones run isolated and don't need external resources so they are always triggered

PR #251, opened 2022-06-06
fix incremental unique array

### LINKED ISSUE #241: Incremental does not work with unique_key that is array
Example model:

    {{ 
      config
      (
        materialized='incremental'
        , unique_key=['A', 'B']
      ) 
    }}
    WITH CTE AS
    (
      SELECT 1 AS A, 2 AS B
      UNION
      SELECT 3 AS A, 4 AS B
    )
    SELECT
      A
      , B
    FROM
      CTE
    

Error:  ('42000', "[42000] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Incorrect syntax near 'using'. (102) (SQLExecDirectW);

Code generated in run/project/model:

    delete from "Test"."dbt"."t"
            using "Test"."dbt"."#t__dbt_tmp"
            where (
                    "Test"."dbt"."#t__dbt_tmp".A = "Test"."dbt"."t".A
                    and
                    "Test"."dbt"."#t__dbt_tmp".B = "Test"."dbt"."t".B
            );

DELETE FROM ... USING is not supported by SQL Server

sdebruyn:
fixes #241

### DIFF
--- a/dbt/include/sqlserver/macros/materializations/models/incremental/merge.sql
+++ b/dbt/include/sqlserver/macros/materializations/models/incremental/merge.sql
@@ -10,7 +10,36 @@
 {% endmacro %}
 
 {% macro sqlserver__get_delete_insert_merge_sql(target, source, unique_key, dest_columns) %}
-  {{ default__get_delete_insert_merge_sql(target, source, unique_key, dest_columns) }};
+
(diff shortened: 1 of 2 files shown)


sdebruyn:
However, the current test seems to succeed ¯\_(ツ)_/¯

Issue #219, opened 2022-05-03
 >Database Error   ('08001', '[08001] [Microsoft][ODBC Driver 18 for SQL Server]SSL Provider: [error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:self signed certificate] (-1) (SQLDriverConnect)')


ogaly-absys:
Hi, I have an error with dbt debug and with the ODBC Driver 18 for SQL Server in my profiles file.
It'think the driver is configured by default with "TrustServerCertificate=Yes" 
it's resolved when i changed the driver to ODBC Driver 17.

dataders:
great callout! I remember reading somewhere that there's a new option you must pass into the connection string for version 18 of the driver like `Something=True`, and now I can't find the doc anywhere! I'll update later when I find it

dataders:
closing in favor of #269

PR #267, opened 2022-07-12
add TrustServerCertificate to Windows login

bsalerno:
`ODBC Driver 18 for SQL Server` gives the following database error when compiling and needs this in the connection string. Tested locally with both Drivers 17 and 18.

Database Error
  ('08001', '[08001] [Microsoft][ODBC Driver 18 for SQL Server]SSL Provider: The certificate chain was issued by an authority that is not trusted.\r\n (-2146893019) (SQLDriverConnect); [08001] [Microsoft][ODBC Driver 18 for SQL Server]Client unable to establish connection (-2146893019); [08001] [Microsoft][ODBC Driver 18 for SQL Server]Invalid connection string attribute (0)')

### DIFF
--- a/dbt/adapters/sqlserver/connections.py
+++ b/dbt/adapters/sqlserver/connections.py
@@ -339,6 +339,7 @@ class SQLServerConnectionManager(SQLConnectionManager):
 
             elif getattr(credentials, "windows_login", False):
                 con_str.append("trusted_connection=yes")
+                con_str.append("TrustServerCertificate=yes")
             elif type_auth == "sql":
                 con_str.append(f"UID={{{credentials.UID}}}")
                 con_str.append(f"PWD={{{credentials.PWD}}}")


dataders:
ah! this solves #219 -- thanks @bsalerno!

note for future -- we should probably start testing using both version 17 and 18 of the driver

sdebruyn:
Thank you @bsalerno, really appreciated! I implemented this in [49e67ed](https://github.com/dbt-msft/dbt-sqlserver/commit/49e67eda47c70fd48196ab0b123bb53d0a64f2e0) so that we are sure that the property is only added to the connection string in a single place (a few lines below).

sdebruyn:
Sorry, coming back on my decision here. I've read more about why this change was introduced in the driver itself and I've reflected this in the default settings for this adapter. Details about how these changes might affect you are in the [CHANGELOG](https://github.com/dbt-msft/dbt-sqlserver/blob/master/CHANGELOG.md) and will be added to the official docs.

Issue #153, opened 2021-07-21
dbt run fails with "The specified schema name userid@domain either does not exist or you do not have permission to use it


jhoolachan:
Hello,

When I create a schema manually, add it to my profiles.yml, and then submit "dbt run", my model is built successfully. However, if the schema specified in my profiles.yml does not exist, the build fails with:

`"The specified schema name 'userid@domain' either does not exist or you do not have permission to use it" "CREATE SCHEMA failed due to previous errors"
`
when active directory integrated authentication is used. This feels like a DB permissions issue but I am able to log into SSMS using active directory integrated and create schemas without issue. I am also not sure why my userid and domain are being concatenated for the schema name when my profiles.yml contains "schema: staging".

I'm running dbt=0.19.2. What additional information can I provide for troubleshooting?

swanderz:
@jhoolachan thanks for putting in the work to get this set up!

more info:
1. what MSFT SQL product are you using? SQL Server (on-prem or IaaS) or Azure SQL?
2. what does a redacted version of your `profiles.yml` look like? It kinda seems like you have your `user` and schema` fields switched.

Full disclosure: I've never personally gotten active directory integrated authentication to work, as my company uses Azure Active Directory, and it is a lot of work to set that up in a demo environment. If you or your company uses is Azure Active Directory, I definitely recommend the `cli` auth method.

jhoolachan:
@swanderz 

I'm using Azure SQL and my profiles.yml is below:

azure-sql-demo:
  target: dev
  outputs:
    dev:
      type: sqlserver
      driver: 'ODBC Driver 17 for SQL Server'
      server: {server}
      port: 1433
      database: {dbt}
      schema: staging
      authentication: ActiveDirectoryIntegrated
      encrypt: true
      
I also see the same issue when using service principal authentication. In that scenario, the compiled schema is : client_id@tenant_id

As noted, both methods work without issue as long as the schema already exists. I have not tried the CLI auth method but can give it a shot.

swanderz:
> I also see the same issue when using service principal authentication. In that scenario, the compiled schema is : client_id@tenant_id

weird! it might be that `'userid@domain'` or `client_id@tenant_id`aren't properly quoted?


I highly recommend the `cli` auth option, then get started after that.

[...]

PR #277, opened 2022-08-11
updates the create schema statement

vigyuv:
Adds authorization based on the create schema syntax defined by Microsoft. 
https://docs.microsoft.com/en-us/sql/t-sql/statements/create-schema-transact-sql?view=sql-server-ver15#remarks

### DIFF
--- a/dbt/include/sqlserver/macros/adapters/schema.sql
+++ b/dbt/include/sqlserver/macros/adapters/schema.sql
@@ -1,9 +1,10 @@
 {% macro sqlserver__create_schema(relation) -%}
+{% auth = config.get("auth") %}
   {% call statement('create_schema') -%}
     USE [{{ relation.database }}];
     IF NOT EXISTS (SELECT * FROM sys.schemas WHERE name = '{{ relation.without_identifier().schema }}')
     BEGIN
-    EXEC('CREATE SCHEMA [{{ relation.without_identifier().schema }}]')
+    EXEC('CREATE SCHEMA [{{ relation.without_identifier().schema }}] {%- if auth -%} AUTHORIZATION: [{{auth}}]')
     END
   {% endcall %}
 {% endmacro %}
@@ -26,6 +27,3 @@
       EXEC('DROP SCHEMA {{ relation.schema }}')
       END  {% endcall %}
 {% endmacro %}
-
-
-{# there is no drop_schema... why? #}


dataders:
@jhoolachan @w0ut0 we think this might solve #153! any chance you want to pip install this branch and see if it works for you? or at least copy the `sqlserver__create_schema()` macro into your project's `macros/` dir and see if it works when you also provide an `auth` config when defining a custom schema...

also perhaps the `auth` field is better defined in the `profiles.yml` because it's not likely that there'd be more than one authorization setting used by a single user?

w0ut0:
Hi @dataders,
thanks for this, and sorry for the delay in testing!
unfortunately this does not solve our issue, I still have to make sure I add the user manually to the database first (he is not implicitly created because his group exists in the DB).
Some logs:
 ``` 
12:42:48.647585 [info ] [MainThread]: Found 302 models, 373 tests, 0 snapshots, 0 analyses, 502 macros, 0 operations, 3 seed files, 73 sources, 2 exposures, 0 metrics
12:42:48.671691 [info ] [MainThread]: 
12:42:48.672493 [debug] [MainThread]: Acquiring new sqlserver connection "master"
12:42:48.674095 [debug] [ThreadPool]: Acquiring new sqlserver connection "list_<dbname>"
12:42:48.688168 [debug] [ThreadPool]: SQLServer adapter: Using sqlserver connection "list_<dbname>".
12:42:48.688431 [debug] [ThreadPool]: SQLServer adapter: On list_<dbname>: USE "<dbname>";
    select  name as [schema]
    from sys.schemas
  
12:42:48.688610 [debug] [ThreadPool]: Opening a new connection, currently in state init
12:42:48.688785 [debug] [ThreadPool]: SQLServer adapter: Using connection string: DRIVER={ODBC Driver 17 for SQL Server};SERVER=dl-<server_id>-sqlserver.database.windows.net,1433;Database=<dbname>;Encrypt=Yes;Application Name=dbt-sqlserver/1.1.0
12:42:49.239272 [debug] [ThreadPool]: SQLServer adapter: Connected to db: <dbname>
12:42:49.243371 [debug] [ThreadPool]: SQLServer adapter: SQL status: OK in 0.55 seconds
12:42:49.246550 [debug] [ThreadPool]: On list_<dbname>: Close
12:42:49.248534 [debug] [ThreadPool]: Acquiring new sqlserver connection "create_<dbname>_dbt_<user_first_name>"
12:42:49.248939 [debug] [ThreadPool]: Acquiring new sqlserver connection "create_<dbname>_dbt_<user_first_name>"
12:42:49.249290 [debug] [ThreadPool]: Creating schema "_ReferenceKey(database='<dbname>', schema='dbt_<user_first_name>', identifier=None)"
12:42:49.259538 [debug] [ThreadPool]: SQLServer adapter: Using sqlserver connection "create_<dbname>_dbt_<user_first_name>".
12:42:49.259777 [debug] [ThreadPool]: SQLServer adapter: On create_<dbname>_dbt_<user_first_name>: USE [<dbname>];
    IF NOT EXISTS (SELECT * FROM sys.schemas WHERE name = 'dbt_<user_first_name>')
    BEGIN
    EXEC('CREATE SCHEMA [dbt_<user_first_name>]')
    END
  
12:42:49.259939 [debug] [ThreadPool]: Opening a new connection, currently in state closed
12:42:49.260118 [debug] [ThreadPool]: SQLServer adapter: Using connection string: DRIVER={ODBC Driver 17 for SQL Server};SERVER=dl-<server_id>-sqlserver.database.windows.net,1433;Database=<dbname>;Encrypt=Yes;Application Name=dbt-sqlserver/1.1.0
12:42:49.325325 [debug] [ThreadPool]: SQLServer adapter: Connected to db: <dbname>
12:42:49.330244 [debug] [ThreadPool]: SQLServer adapter: SQL status: OK in 0.07 seconds
12:42:49.331522 [debug] [ThreadPool]: SQLServer adapter: Database error: ('42000', "[42000] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Cannot find the user '', because it does not exist or you do not have permission. (15151) (SQLMoreResults)")
12:42:49.331775 [debug] [ThreadPool]: On create_<dbname>_dbt_<user_first_name>: ROLLBACK
12:42:49.331944 [debug] [ThreadPool]: On create_<dbname>_dbt_<user_first_name>: Close
12:42:49.333972 [debug] [MainThread]: Connection 'master' was properly closed.
12:42:49.334234 [debug] [MainThread]: Connection 'create_<dbname>_dbt_<user_first_name>' was properly closed.
```
The only thing I see that would help is creating the user if it doesn't exist yet on schema creation, but I'm not sure that is something we want.

sdebruyn:
I think we need something like this. Right now everything is always owned by `dbo`. You're using a configuration variable, can we make it more seamless in some cases? E.g. if the user uses SQL authentication, shouldn't we just use SQL username?

sdebruyn:
I ran into the same issue but then on Synapse at a customer a few months ago. We ended up giving the users more permissions. The nasty part about this is the following part:

You can give permissions to AAD groups if you create those as external database principals. We added the auto provisioning feature to the adapter to make that part of the flow nicer.
But when you are signed in as a user of an AAD group and you don't have the right roles, you cannot create schemas. Since, implicitly, it will try to link the schema to a user which it has to create, but it does not create AAD users.

What is the key missing piece here is that Microsoft should allow us to let AAD groups own schemas, then the default would be okay and this PR would not even be needed.

I don't think we should provide a solution for this inside the adapter as there is no way of knowing (from the side of the adapter) what your AAD username is. (CURRENT_USER gives the name of your group, or - if you have the db_owner role - it just gives `dbo`).

In our case we had to give the users the CONTROL on the database, we should further research how we can create and use schemas without such grants.

sdebruyn:
I am going to rebase this PR, rename the config setting to make it clearer what it does and add tests for it.

sdebruyn:
This doesn't seem to work, the config is ignored at the stage the schema is created

sdebruyn:
Closing in favor of #382

PR #286, opened 2022-09-14
Automatically provision Azure AD users if enabled

sdebruyn:
This would automatically provision Azure AD users used in grants if they are not created in the database yet. This is using contained database users which would only be possible in Azure SQL and Azure Synapse.

### DIFF
--- a/dbt/include/sqlserver/macros/adapters/apply_grants.sql
+++ b/dbt/include/sqlserver/macros/adapters/apply_grants.sql
@@ -9,9 +9,6 @@
 {% endmacro %}
 
 
-
-
-
 {%- macro sqlserver__get_grant_sql(relation, privilege, grantees) -%}
     {%- set grantees_safe = [] -%}
     {%- for grantee in grantees -%}
--- a/dbt/include/sqlserver/macros/adapters/apply_grants.sql
+++ b/dbt/include/sqlserver/macros/adapters/apply_grants.sql
@@ -9,41 +9,33 @@
 {% endmacro %}
 
 
-{%- macro sqlserver__get_dcl_statement_list(relation, grant_config, get_dcl_macro) -%}
-    {#
-      -- Unpack grant_config into specific privileges and the set of users who need them granted/revoked.
-      -- Depending on whether this database supports multiple grantees per statement, pass in the list of
-      -- all grantees per privilege, or (if not) template one statement per privilege-grantee pair.
-      -- `get_dcl_macro` will be either `get_grant_sql` or `get_revoke_sql`
-    #}
-    {%- set dcl_statements = [] -%}
-    {%- for privilege, grantees in grant_config.items() %}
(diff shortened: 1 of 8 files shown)


JCZuurmond:
LGTM; Do you have a link to the docs of `sqlserver_get_grant_sql(relation, privilege, grantees)`? To know what the parameters are

dataders (changes requested):
i love the auto user-creation feature. however, there looks to be too much happening inside of `sqlserver_get_grant_sql`. It's only supposed to create the GRANT statements 

perhaps the user creation logic would fit better into the [apply_grants](https://cs.github.com/dbt-labs/dbt-core/blob/f5a94fc77428426002ef2a908ebf57a32597a627/core/dbt/include/global_project/macros/adapters/apply_grants.sql#L139) macro? might make sense for this overridden version to be just on dbt-sqlserver first, then later we might be able to add a new conditional flow to the default__ version in dbt-core. i imagine this feature could be useful to others

dataders:
> LGTM; Do you have a link to the docs of `sqlserver_get_grant_sql(relation, privilege, grantees)`? To know what the parameters are

@JCZuurmond check out the [global_project's apply_grants.sql](https://github.com/dbt-labs/dbt-core/blob/f5a94fc77428426002ef2a908ebf57a32597a627/core/dbt/include/global_project/macros/adapters/apply_grants.sql#L139) as well as the grant section of the 1.2 upgrade guide (https://github.com/dbt-labs/dbt-core/discussions/5468)

sdebruyn:
> perhaps the user creation logic would fit better into the [apply_grants](https://cs.github.com/dbt-labs/dbt-core/blob/f5a94fc77428426002ef2a908ebf57a32597a627/core/dbt/include/global_project/macros/adapters/apply_grants.sql#L139) macro? might make sense for this overridden version to be just on dbt-sqlserver first, then later we might be able to add a new conditional flow to the default__ version in dbt-core. i imagine this feature could be useful to others

@dataders Thank you for your review! Agree with the above. I changed it so that it better follows the intentions of the other macros in that file.

dbt/include/sqlserver/macros/adapters/apply_grants.sql
```diff
@@ -6,4 +6,81 @@
...
+          {% endfor -%}
+        {%- endif -%}
+    {%- endfor -%}
+    {{ return(dcl_statements) }}
+{%- endmacro %}
+
+
+{% macro get_provision_sql(relation, privilege, grantees) %}
+    {% for grantee in grantees %}
+        if not exists(select name from sys.database_principals where name = '{{ grantee }}')
```
sdebruyn, line 45: This check currently fails since it does the comparison on the safe user

sdebruyn: fixed

PR #294, opened 2022-09-19
UUID for Temp Table

### LINKED ISSUE #293: get_columns_in_relation() returns wrong columns for Temp Relations with the same relation identifier (table) as other schemas
```
2022-09-15 17:57:11.594602 (Thread-4): On model.mart.DataMart__AzureAD__USERS: 
      SELECT
          column_name,
          data_type,
          character_maximum_length,
          numeric_precision,
          numeric_scale
      FROM
          (select
              ordinal_position,
              column_name,
              data_type,
              character_maximum_length,
              numeric_precision,
              numeric_scale
          from [DataMart].INFORMATION_SCHEMA.COLUMNS
          where table_name = '#USERS__dbt_tmp'
            and table_schema = 'AzureAD'
          UNION ALL
          select
              ordinal_position,
              column_name collate database_default,
              data_type collate database_default,
              character_maximum_length,
              numeric_precision,
              numeric_scale
          from tempdb.INFORMATION_SCHEMA.COLUMNS
          where table_name like '#USERS__dbt_tmp%') cols
      order by ordinal_position
```
      
This sql comes from https://github.com/dbt-msft/dbt-sqlserver/blob/master/dbt/include/sqlserver/macros/adapters/columns.sql#L27
    
My issue is that I have a schema `AzrueAD` with a table `Users` and another Schema with a table `Users` so I get columns from the wrong table in the relation.

visch:
Closes #293

### DIFF
--- a/dbt/include/sqlserver/macros/adapters/relation.sql
+++ b/dbt/include/sqlserver/macros/adapters/relation.sql
@@ -1,5 +1,6 @@
 {% macro sqlserver__make_temp_relation(base_relation, suffix) %}
-    {% set tmp_identifier = '#' ~  base_relation.identifier ~ suffix %}
+    {% set uuid = run_query("SELECT REPLACE(NEWID(), '-','_') ")[0].values()[0] %}
+    {% set tmp_identifier = '#' ~  base_relation.identifier ~ uuid ~ suffix %}
     {% set tmp_relation = base_relation.incorporate(
                                 path={"identifier": tmp_identifier}) -%}
 


visch:
Not sure if you all want to do this or not, but it did fix my pretty niche issue.

dbt/include/sqlserver/macros/adapters/relation.sql
```diff
@@ -1,5 +1,6 @@
 {% macro sqlserver__make_temp_relation(base_relation, suffix) %}
-    {% set tmp_identifier = '#' ~  base_relation.identifier ~ suffix %}
+    {% set uuid = run_query("SELECT REPLACE(NEWID(), '-','_') ")[0].values()[0] %}
```
sdebruyn, line 2: I'd prefer to generate this ID in Jinja instead of in SQL

sdebruyn (changes requested):

dbt/include/sqlserver/macros/adapters/relation.sql
visch: https://gist.github.com/hijonathan/5db3d8ec64e267d06c32 if someone wants to take a stab at it, if I have time I"ll come back to this

SportsTribution: How about
{%- set uuid = local_md5(base_relation.schema) -%}
?

sdebruyn:
I don't think we need this anymore as #368 should fix this issue and is more consistent with what is expected in SQL Server

PR #254, opened 2022-06-08
fix: adds pyodbc output converter for DATETIMEOFFSET

baldwicc:
Fixes #253

Refs:
https://github.com/mkleehammer/pyodbc/wiki/Using-an-Output-Converter-function
https://github.com/mkleehammer/pyodbc/issues/134#issuecomment-281739794

PR #346, opened 2023-04-13
fix: adds pyodbc output converter for DATETIMEOFFSET data types

### LINKED ISSUE #253: Cannot use `dbt source freshness` with DATETIMEOFFSET column types
Running into issues with [`dbt source freshness`](https://docs.getdbt.com/reference/resource-properties/freshness) pointing at DATETIMEOFFSET columns, with pyodbc throwing the following error:

````
ODBC data type -155 is not supported.
````

Directly relates to https://github.com/mkleehammer/pyodbc/issues/134

baldwicc:
Fixes #253, supercedes #254

pyodbc is unlikley to include a handler for DATETIMEOFFSET data types - see [mkleehammer/pyodbc#134](https://github.com/mkleehammer/pyodbc/issues/134).

So, this PR adds support to convert these datatypes to standard `datetime` objects based on public MS documentation.

This should allow [freshness](https://docs.getdbt.com/reference/resource-properties/freshness) functonality to use DATETIMEOFFSET columns without needing to use CAST/CONVERT in [loaded_at_field](https://docs.getdbt.com/reference/resource-properties/freshness#loaded_at_field) resource properties.

Tests are against `str(value: datetime)` default string format, since this asserts compatibility with:
- [innoverio.vscode-dbt-power-user](https://github.com/innoverio/vscode-dbt-power-user)'s query preview
- exotic use of the [run_query](https://docs.getdbt.com/reference/dbt-jinja-functions/run_query) jinja macro

### DIFF
--- a/dbt/adapters/sqlserver/sql_server_connection_manager.py
+++ b/dbt/adapters/sqlserver/sql_server_connection_manager.py
@@ -1,5 +1,6 @@
 import struct
 import time
+import datetime as dt
 from contextlib import contextmanager
 from itertools import chain, repeat
 from typing import Any, Callable, Dict, Mapping, Optional, Tuple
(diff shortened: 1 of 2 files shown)


sdebruyn:
@baldwicc Thank you! Sorry for the long wait, merging this today

PR #356, opened 2023-05-12
bugfix: Install a correct version of `dbt-core` that is compatible with the library

### LINKED ISSUE #348: setup.py includes the wrong requirement for dbt-core
As of `dbt-sqlserver` v1.3.1, `setup.py` includes the following lines:

> install_requires=[
>      "dbt-core>=1.3.0"
> ],

This means that when installing `dbt-sqlserver` using any package manager the wrong version of `dbt-core` will always be installed. This currently leads to incompability issues since `dbt-core` is at v1.5.0 but only v1.3.0 is supported. Users should expect to see the following errors when using `dbt debug` and `dbt run`:

> module 'dbt.exceptions' has no attribute 'DatabaseException'

To fix, `setup.py` should be changed to:

> install_requires=[
>      "dbt-core==1.3.0"
> ],

l-j:
This is a bugfix 1.3.2 release, it resolves #348.

As we already have an ongoing v1.4 release on master, this PR requires a v1.3 branch to be created from the v.1.3.1 tag, which I do not think I can do myself. @dataders @sdebruyn can you help with this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+### v1.3.2
+
+#### Fixes
+
+* Install a correct version of `dbt-core` that is compatible with the library 
+
+
 ### v1.3.0
 
 #### Features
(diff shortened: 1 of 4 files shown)


dataders:
I just made a branch, `v1.3.latest`, based of of the v1.3.1 release tag. Can you open your changes against that branch?

PR #357, opened 2023-05-12
Install a correct version of `dbt-core` that is compatible with the library

l-j:


l-j:
Duplicates https://github.com/dbt-msft/dbt-sqlserver/pull/357

PR #368, opened 2023-05-14
Fix concurrency issues and document create as

sdebruyn:
- should fix all the concurrency issues we currently experience (deadlocks)
- stick a bit closer to the upstream/core macros where possible

### DIFF
--- a/dbt/include/sqlserver/macros/adapters/columns.sql
+++ b/dbt/include/sqlserver/macros/adapters/columns.sql
@@ -1,39 +1,19 @@
 {% macro sqlserver__get_columns_in_relation(relation) -%}
   {% call statement('get_columns_in_relation', fetch_result=True) %}
 
-    with
-        regular_db_cols as (
-            select
-                ordinal_position,
-                column_name,
-                data_type,
-                character_maximum_length,
-                numeric_precision,
-                numeric_scale
-            from [{{ relation.database }}].INFORMATION_SCHEMA.COLUMNS
-            where table_name = '{{ relation.identifier }}'
-              and table_schema = '{{ relation.schema }}'
-        ),
-
-        temp_db_cols as (
-            select
-                ordinal_position,
-                column_name collate database_default as column_name,
-                data_type collate database_default as data_type,
-                character_maximum_length,
-                numeric_precision,
-                numeric_scale
-            from tempdb.INFORMATION_SCHEMA.COLUMNS
-            where table_name like '{{ relation.identifier }}%'
-        ),
-
-        all_cols as (
-            select *
-            from regular_db_cols
-            union
-            select *
-            from temp_db_cols
-        )
+    with mapping as (
+        select
+            row_number() over (partition by object_name(c.object_id) order by c.column_id) as ordinal_position,
+            c.name collate database_default as column_name,
+            t.name as data_type,
+            c.max_length as character_maximum_length,
+            c.precision as numeric_precision,
+            c.scale as numeric_scale
+        from [{{ 'tempdb' if '#' in relation.identifier else relation.database }}].sys.columns c
+        inner join sys.types t
+        on c.user_type_id = t.user_type_id
+        where c.object_id = object_id('{{ 'tempdb..' ~ relation.include(database=false, schema=false) if '#' in relation.identifier else relation }}')
+    )
 
     select
         column_name,
@@ -41,8 +21,7 @@
         character_maximum_length,
         numeric_precision,
         numeric_scale
-    from
-        all_cols
+    from mapping
     order by ordinal_position
 
   {% endcall %}
--- a/dbt/include/sqlserver/macros/materializations/models/table/create_table_as.sql
+++ b/dbt/include/sqlserver/macros/materializations/models/table/create_table_as.sql
@@ -1,26 +1,43 @@
 {% macro sqlserver__create_table_as(temporary, relation, sql) -%}
+   {#- TODO: add contracts here when in dbt 1.5 -#}
+   {%- set sql_header = config.get('sql_header', none) -%}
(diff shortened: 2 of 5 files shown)


dbt/include/sqlserver/macros/adapters/columns.sql
```diff
@@ -1,48 +1,27 @@
...
+            c.name collate database_default as column_name,
+            t.name as data_type,
+            c.max_length as character_maximum_length,
+            c.precision as numeric_precision,
+            c.scale as numeric_scale
+        from [{{ 'tempdb' if '#' in relation.identifier else relation.database }}].sys.columns c
+        inner join sys.types t
+        on c.user_type_id = t.user_type_id
+        where c.object_id = object_id('{{ 'tempdb..' ~ relation.include(database=false, schema=false) if '#' in relation.identifier else relation }}')
+    )
```
dataders, line 16: can you help me understand what this does differently than before

dbt/include/sqlserver/macros/materializations/models/table/create_table_as.sql
```diff
@@ -1,26 +1,35 @@
 {% macro sqlserver__create_table_as(temporary, relation, sql) -%}
+   {#- TODO: add contracts here when in dbt 1.5 -#}
+   {%- set sql_header = config.get('sql_header', none) -%}
    {%- set as_columnstore = config.get('as_columnstore', default=true) -%}
-   {% set tmp_relation = relation.incorporate(
-   path={"identifier": relation.identifier.replace("#", "") ~ '_temp_view'},
-   type='view')-%}
    {%- set temp_view_sql = sql.replace("'", "''") -%}
+   {%- set tmp_relation = relation.incorporate(
```
dataders, line 6: if we're using temp tables correctly, shouldn't this be the below?
when is `.replace("#", "")` necessary?
also 
```suggestion
 {% set tmp_relation = make_temp_relation(relation,'_temp_view')
```

dbt/include/sqlserver/macros/materializations/models/table/create_table_as.sql
```diff
@@ -1,26 +1,35 @@
...
     {{ temp_view_sql }}
     ');
 
-   SELECT * INTO {{ relation }} FROM
-    {{ tmp_relation }}
+   -- select into the table and create it that way
+   {# TempDB schema is ignored, always goes to dbo #}
+   SELECT *
+   INTO {{ relation.include(database=False, schema=(not temporary))  }}
+   FROM {{ tmp_relation }}
```
dataders, line 25: the mechanics of this are also interesting to me. is the order of operations around this
1. create `tmp_relation` which is a view
2. (this step) materialized this data into `dbo` schema
3. rename the resulting table into the correct database and schema?

dbt/include/sqlserver/macros/adapters/columns.sql
sdebruyn: When you create a temp table, SQL Server will actually name it `#yourname_____...____somerandomvalue`
When you reference the temp table by its regular name (`#yourname`), you can find it and use it as normal. In this case, the previous version looked for the temp table using information_schema (ansi/cross-db version of sys.*) but it looked for all matches with `LIKE "#yourname%`. In a single-user/single-session scenario, you'd only get a single result. But if multiple users are working on the same database (with shared tempdb) and use temp tables with the same name (as temp tables don't have a schema), you'd get multiple matches. This is why sometimes our test failed when multiple instances were running at the same time, the result was not reliable.
This only reliable way to get all columns in *just your* temp table, without including the ones from other users is by referencing it in a normal way and let SQL Server do its magic to find back the one with the name suffix. In our case this means that you have to ref it using an object_id which is always unique. But you can't use object_id with information_schema, so I had to revert to sys.columns and sys.types to get the same information.

dbt/include/sqlserver/macros/materializations/models/table/create_table_as.sql
sdebruyn: The operations here are there because SQL Server doesn't have CTAS (which Synapse does, so to be overridden there again). The closest thing in SQL Server is SELECT ... INTO target FROM source. But the source in the FROM clause there, cannot be a subquery. So this first creates a temporary view to serve as the FROM source as the view is the only thing which can just take the SQL as-is. The temporary view is dropped again at the end.

dbt/include/sqlserver/macros/materializations/models/table/create_table_as.sql
sdebruyn: This is doing it the other way around. create_table_as is called to create temporary tables when working with snapshots and incremental models. In that case the `temporary` argument is set to `true`. In Postgres, this just means you have to add the keyword `temporary` in your CREATE TABLE syntax, but in SQL Server we have to prefix with `#`. This happens in `sqlserver__make_temp_relation` in relation.sql. In that case we will receive a relation identifier here which already has the `#` prefix. But for our temp view (see other comment below), we cannot have the `#` prefix in there. So that's removed here so that we can continue with business as usual.

dbt/include/sqlserver/macros/adapters/columns.sql
sdebruyn: 1. Seems a bit faster when running multiple tests concurrently.
2. My assumption is that the override is no longer needed.

dbt/include/sqlserver/macros/materializations/models/table/create_table_as.sql
sdebruyn: The database=False is because we use USE [database] right above. The schema=(not temporary) is because tempdb only has dbo and will still use dbo if you put something else. So it would be confusing if you'd dive into the logs and then find out that the schema you've used doesn't exist (guess who had to do this).

PR #370, opened 2023-05-15
remove azure max parallel test runs

sdebruyn:


sdebruyn:
To be merged after https://github.com/dbt-msft/dbt-sqlserver/pull/369

PR #386, opened 2023-05-17
use MD5 hashing to avoid long index names

### LINKED ISSUE #317: create_nonclustered_index macro does not check for index name length
I get the following error:
_Failed to execute query. Error: The identifier that starts with 'test__index_on_person_id_includes_l1234567_source_value_gender_concept_id_l_source_collection_elt_loaded_at_race_conce' is too long. Maximum length is 128._

When executing:

{{
    config({
        "materialized": 'table',
        "as_columnstore": False,
        "post-hook": [
                "{{ create_nonclustered_index(columns = ['person_id'], includes = ['l1234567_id_source_value', 'gender_concept_id','l1234567_source_collection','elt_loaded_at','race_concept_id','hospital_id_source_value' ]) }}",                                                        
        ]
    })
}}

The create_nonclustered_index  macro does not check the length of the index name:
  `{% set idx_name = (this.table + '__index_on_' + columns|join('_')|replace(" ", "_") + '_includes_' + includes|join('_')|replace(" ", "_")) %}`

So when there are too many included columns, the index name is too long.

sdebruyn:
resolves #317

### DIFF
--- a/dbt/include/sqlserver/macros/adapters/indexes.sql
+++ b/dbt/include/sqlserver/macros/adapters/indexes.sql
@@ -134,7 +134,7 @@ select @drop_remaining_indexes_last = (
 
 {{ log("Creating clustered index...") }}
 
-{% set idx_name = this.table + '__clustered_index_on_' + columns|join('_') %}
+{% set idx_name = "clustered_" + local_md5(columns | join("_")) %}
 
 if not exists(select *
                 from sys.indexes with (nolock)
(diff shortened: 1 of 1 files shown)


NicolasPA:
This index naming change is a breaking change which causes errors on our DBT models that create clustered index:
`Cannot create more than one clustered index on table *. Drop the existing clustered index * before creating another. (1902)`

Additionally, this would silently create duplicated indexes for non-clustered indexes because they can be more than one.

The easy is workaround is to full-refresh the DBT models, but it may be computationally costly.
The laborious but computationally efficient workaround is to manually rename all the indexes according to the new naming scheme.

Maybe it could be marked as a breaking-change in the[ v1.4.1 release note](https://github.com/dbt-msft/dbt-sqlserver/releases/tag/v1.4.1) to help users who will get the errors.

Issue #407, opened 2023-05-26
SQL Server error 8003 "too many parameters" on large seed with dbt-sqlserver >= 1.4.1


NicolasPA:
From `dbt-sqlserver` >= 1.4.1, when running one of our big seed, we're getting the SQL Server 2016 error 8003:
> The incoming request has too many parameters. The server supports a maximum of 2100 parameters. Reduce the number of parameters and resend the request.

Would there be a query generated by DBT with a very large `where` condition?

The failing seed is 2.2 MB, 10 columns x 17k rows, including 3 large columns around `varchar(250)`.
We have another smaller, but still big, seed that woks just fine with 1.4 MB, 5 columns x 24k rows.

For now, we're stuck with  `dbt-sqlserver` ==1.3.2 because 1.4.0 has issues with indexes and versions above have the issue described above with the seed.

Despite this issue, it's been a great experience working with `dbt-sqlserver`, thanks for your work!

NicolasPA:
Thank you for pointing out the batch size config!
I will try when I can.

NicolasPA:
Sorry to come back so late, we tested overwriting the macro and it indeed worked, thank you !
We also tried setting [`max_batch_size`](https://docs.getdbt.com/reference/resource-configs/mssql-configs#seeds) but that didn't work.

I think it would still require a systematic solution to avoid this workaround.

lucaslortiz:
There are two issues here:
1) The [max_batch_size](https://docs.getdbt.com/reference/resource-configs/mssql-configs#seeds) var is not working

2) Even though this workaround worked in your case, it is not the final solution. The size of the batch is calculated based on the maximum number of parameters allowed, as you can see below. The issue is that it considered 2100 to be the maximum, while the real value is 2098 ([reference here](https://stackoverflow.com/questions/8050091/sqlcommand-maximum-parameters-exception-at-2099-parameters)). In some specific cases, it will still fail, even if you set a lower number for the batch - for example, if you have 21 columns and more than 2100 rows, with a batch of 100.

[Calculated maximum number of parameters](https://github.com/dbt-msft/dbt-sqlserver/blob/master/dbt/include/sqlserver/macros/materializations/seeds/helpers.sql)
![image](https://github.com/dbt-msft/dbt-sqlserver/assets/82912062/ebc79e39-3135-4db3-b100-ba880b28f161)

I'll create a PR shortly.

joshmarkovic:
[...]

PR #438, opened 2023-08-21
Fix calc_batch_size maximum number of parameters allowed

lucaslortiz:
Fixing issue #407

The correct maximum number of parameters is 2098, not 2100, as previously set.
Reference [here](https://stackoverflow.com/questions/8050091/sqlcommand-maximum-parameters-exception-at-2099-parameters).


matsonj:
@sdebruyn can we get a review on this one? Good work here by Lucas to get this fixed and its a regression vs 1.3

PR #477, opened 2024-02-15
Fix for adding columns to snapshots

ericmuijsvanoord:
Fix to be able to add columns in snapshots.

Changes

1. column_name > column.quoted (column_name does not exist, also in fabric)
2. Changed into statement such that columns appear after SELECT *

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,11 +1,21 @@
 {
-	"name": "Python 3",
-	"image": "mcr.microsoft.com/devcontainers/python:1-3.10-bookworm",
-	"features": {
-		"ghcr.io/devcontainers/features/docker-in-docker:2": {}
-	},
-	"postStartCommand": "bash .devcontainer/setup_odbc.sh && bash .devcontainer/install_pyenv.sh && bash .devcontainer/setup_env.sh",
-	"runArgs": [
-        "--env-file", "${localWorkspaceFolder}/.devcontainer/test.env"
-    ]
+    "name": "Python 3",
+    "image": "mcr.microsoft.com/devcontainers/python:1-3.10-bookworm",
+    "features": {
+        "ghcr.io/devcontainers/features/docker-in-docker:2": {}
+    },
+    "postStartCommand": "sudo bash .devcontainer/setup_odbc.sh && bash .devcontainer/install_pyenv.sh && bash .devcontainer/setup_env.sh",
+    "containerEnv": {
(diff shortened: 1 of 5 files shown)


ericmuijsvanoord:
@schlich Let me know If I need to do more :)

schlich:
Thanks for the PR!  Ideally I'd like to see a corresponding test that captures the bug -- The fix makes enough sense to me.  Let's try to replicate this failure in a test:

> I get an error when I add new columns to an existing snapshot (destination is available).

I'm not too familiar with snapshots so if you feel like you're not sure how to implement the test in the code base, let's discuss a high level approach here and figure out how it translates.

ericmuijsvanoord:
I don't have a full dev environment for this yet so might take a while.
High level approach:

1. Create a source table
2. Snapshot the table
3. Add column to source table (varchar(100)) for example using ALTER TABLE source_table ADD COLUMN new_varchar varchar(100) NULL
4. Snapshot again
5. Validate if column is added tot the snapshot table

ericmuijsvanoord:
@cody-scott @schlich 
Since this is a bug fix of sqlserver adapter, why can't we reuse testing from dbt-core?
Snapshots are part of the functionalities of dbt-core and covered in these tests. Do we run all tests in dbt-core as well for better coverage?

cody-scott:
Adding my two cents. 

Yes if there is a core test(s) from the default test suite then those should get implemented and overridden as needed.

sdebruyn:
Chiming in here: my colleagues and myself recently did a lot of work for dbt-synapse and dbt-fabric (bringing both to dbt 1.8). We also noticed that dbt-fabric does not implement all the tests it should. It has all tests until dbt 1.4 (that's the point until where I implemented all of them and transferred ownership to Microsoft) but all newer ones haven't been added yet. Our assignment focussed solely on dbt-synapse so we implemented them there. You might be able to leverage a lot of the work we did there.

ericmuijsvanoord:
Thanks for the comments @sdebruyn @cody-scott 

See the latest commits in which I overwrite some of the snapshot test functions to get it to work well.

The main branch did not work as expected:

image

ericmuijsvanoord:
> eagues and myself recently did a lot of work for dbt-synapse and dbt-fabric (bringing both to dbt 1.8). We also noticed that dbt-fabric does not implement all the tests it shou

Maybe a new issue / enhancement should be raised to solve this.

ericmuijsvanoord:
@schlich : let me know in case I need to update something

PR #504, opened 2024-06-24
Simplify for scalability

axellpadilla:
This adapater is for sql-server, add syntax works on sql server, but not on fabric, its expensive to do drop and recreate big tables


cody-scott:
@axellpadilla can you take a look at the 1.8.0rc1 and see if that address this issue?

axellpadilla:
@cody-scott from my understanding of the changes yes, need to merge and deploy it to my local to test it but looks ok, closed because any problem should patched on the new 1.8.0rc1

PR #541, opened 2024-09-09
540 units tests not working as expected

cody-scott:
Adds logic for unit tests supported in 1.8


cody-scott:
Unit tests all pass standard DBT suite. Merging in and will resolve new issues as they arise.

PR #568, opened 2024-10-07
Update unit_tests.sql to replace square brackets with double quotes

danakee:
Replace square brackets with double quotes and insert "." between schema and view in setting of test_view and expected_view.

Result was [dbo.testview_123] will now be "dbo"."testview_123" Result was [dbo.expectedview_123] will now be "dbo"."expectedview_123"


axellpadilla:
superseded by #576

PR #576, opened 2024-10-24
Update the derivation of the test view name

### LINKED ISSUE #575: Tests fail due to collisions in view names
Due to the usage of the selection of a random number from a range in instances when a large number of tests are being run I will occasionally run into view name collisions which cause the test to fail.

Added potential fix to generate a more unique view name using the test SQL itself

Benjamin-Knight:
Resolves #575 Use the local MD5 macro to create a more unique hash based on the SQL of the test itself

### DIFF
--- a/dbt/include/sqlserver/macros/materializations/tests.sql
+++ b/dbt/include/sqlserver/macros/materializations/tests.sql
@@ -8,7 +8,7 @@
   END
 
   {% set testview %}
-    [{{ target.schema }}.testview_{{ local_md5(main_sql) }}]
+    [{{ target.schema }}.testview_{{ local_md5(main_sql) }}_{{ range(1300, 19000) | random }}]
   {% endset %}
 
   {% set sql = main_sql.replace("'", "''")%}
--- a/dbt/include/sqlserver/macros/materializations/tests.sql
+++ b/dbt/include/sqlserver/macros/materializations/tests.sql
@@ -8,7 +8,7 @@
   END
 
   {% set testview %}
-    [{{ target.schema }}.testview_{{ local_md5(main_sql) }}_{{ range(1300, 19000) | random }}]
+    [{{ target.schema }}].[testview_{{ local_md5(main_sql) }}_{{ range(1300, 19000) | random }}]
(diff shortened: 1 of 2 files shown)


cody-scott:
Thanks @Benjamin-Knight. Will set some time aside this week to review.

johannes-becker-bt:
Hi,
we're having the same problem.
**I'm fine with the solution and this can be ignored if it causes too much trouble** 
Isn't the more common approach to hash the "dbt name" of the test (which should be unique as well)? Or hash plus a random seed in case someone somehow manages to be really crazy with column and table names.
You could probably have two tests that run the same sql but yaml logic should forbid that there is the same test twice.


Example, so one would hash relationships_deac_deadline_availability_conflict_dh_mttr_matter_id__dh_mttr_id__ref_mttr_matter_ instead of the underlying SQL.
image
image

Cheers
Hannes

johannes-becker-bt:
Only if you feel like it. 
Main reason I would stick with your solution: we got main_sql available. All other solutions would need much more tinkering in different places (Unless the test name is somewhere in target or something, I'm never sure about that...)

johannes-becker-bt:
Merry Christmas!
I tested it with a local macro and it works like a charm.
Also it might make sense to combine the solution with the other change regarding tests
https://github.com/dbt-msft/dbt-sqlserver/pull/586
Have a great New Year!
Hannes

timestescrane:
^ Yes please! Would love to hit 2 birds with one stone on this

timestescrane:
@cody-scott would you be able to give this a look-over this week? Would help unblock some upstream work, well, at work. :)

timestescrane:
@Benjamin-Knight would you be able to merge this PR?

Benjamin-Knight:
I do not have permissions to do that, all checks are now passing that the failing tests were updated via #589 so just waiting on @cody-scott to sign it off now its all passing.

timestescrane:
@cody-scott would you be able to merge this PR? Getting to a point where having this fix in is going to make my life a lot better at work.

PR #582, opened 2024-12-02
Fix broken imports from Fabric

cody-scott:
dbt-fabric has updated the method `get_pyodbc_attrs_before` to `get_pyodbc_attrs_before_credentials` and added `get_pyodbc_attrs_before_accesstoken` in version 1.8.8.

This updates for those new changes.

### DIFF
--- a/dbt/adapters/sqlserver/__version__.py
+++ b/dbt/adapters/sqlserver/__version__.py
@@ -1,1 +1,1 @@
-version = "1.8.4"
+version = "1.8.5"


Benjamin-Knight:
Do we need to update the dependencies of version 1.8.5 to be Fabric 1.8.8 and above? Right now its set to the 1.8 or above which caused an error when I updated to version 1.8.5 as it left my Fabric at 1.8.7 which lacks the exposed API this update is implementing.

PR #587, opened 2024-12-31
Update README.md

NandanHegde15:


axellpadilla:
dbt-sqlserver supports SQL Server and Azure SQL services. SQL database in Microsoft Fabric may be compatible because it uses the Azure SQL Database engine, but it should only be documented as supported after adapter-level testing.

PR #596, opened 2025-02-11
Change dbt upperbound to 2.0

ataft:
As of dbt 1.8, adapter versions no longer need to match the dbt-core version. By the rules of semantic versioning, dbt-core should not introduce breaking changes in a minor or patch, so this should be safe until dbt-core 2.0.

https://docs.getdbt.com/docs/dbt-versions/core#how-we-version-adapter-plugins


axellpadilla:
closed with #598

PR #598, opened 2025-02-18
Update to 1.9

### LINKED ISSUE #597: Bump to version 1.9 to track Fabric Adapter
The upstream fabric adapter has been bumped to version 1.9 to support micro batching, looking at the code there appears to be no reason we cannot just track this update as all the requirements have been implemented in a way that this adapter will just run.

I did note that they have removed the merge incremental model as a supported type. This could be a breaking change if someone was using this type.

Benjamin-Knight:
Closes #597.

All the tests appear to pass and I've successfully used the micro batch functionality locally using this version of the adapter. If anyone can think of additional tests that we require I can look at adding them but the Fabric adapter seemed to have the new functionality covered.

### DIFF
--- a/.github/workflows/integration-tests-sqlserver.yml
+++ b/.github/workflows/integration-tests-sqlserver.yml
@@ -18,7 +18,7 @@ jobs:
     name: Regular
     strategy:
       matrix:
-        python_version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python_version: ["3.9", "3.10", "3.11", "3.12"]
         msodbc_version: ["17", "18"]
         sqlserver_version: ["2017", "2019", "2022"]
         collation: ["SQL_Latin1_General_CP1_CS_AS", "SQL_Latin1_General_CP1_CI_AS"]
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -12,7 +12,7 @@ jobs:
   publish-docker-client:
     strategy:
       matrix:
-        python_version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
+        python_version: ["3.9", "3.10", "3.11", "3.12"]
         docker_target: ["msodbc17", "msodbc18"]
     runs-on: ubuntu-latest
     permissions:
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -18,7 +18,7 @@ jobs:
     name: Unit tests
     strategy:
       matrix:
-        python_version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python_version: ["3.9", "3.10", "3.11", "3.12"]
     runs-on: ubuntu-latest
     permissions:
       contents: read
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,7 +32,7 @@ repos:
     hooks:
       - id: absolufy-imports
   - repo: 'https://github.com/hadialqattan/pycln'
-    rev: v2.4.0
+    rev: v2.5.0
     hooks:
       - id: pycln
         args:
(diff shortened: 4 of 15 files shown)


Benjamin-Knight:
I'm not getting the requirements issue locally and looking it any dbt-core 1.9 or higher should satisfy pip so not quite sure what is causing the build issue.

I can add an extra requirement to guide pip into picking the correct dbt-core version unless someone knows exactly what's going wrong on has a better idea of how to resolve?

cody-scott:
@Benjamin-Knight  can you update the `setup.py` file to unpin 
```python
        "dbt-fabric<1.10.0",
        "dbt-core<1.10.0",
```

and the `dev_requirements.txt` file 
```
dbt-tests-adapter
```

I think unpinning those two should pass the test to start

Benjamin-Knight:
Should I add back in a dbt-tests-adapter adapter version so that it doesn't try and install something for a lower version of dbt-core?

Benjamin-Knight:
@cody-scott I'd be tempted to pin the version of dbt-fabric to 1.9.3, I've got concerns they could update a macro to support nested CTEs which would then break this adapter for anyone who installed it after they release that code.

cody-scott:
I'm not opposed to that either. There is a scheduled job that runs the tests on sundays to check if any fabric merges breaks this adapter. I could make it nightly though if that's better.

Benjamin-Knight:
The DBT documentation states
Assuming you adapter plugin is pinned to a specific minor version of dbt-core (e.g. ~=1.1.0), you can use the same pin for dbt-tests-adapter.
So I'm going to try that in the requirements-dev and see if that resolves it as its trying to install an old version on the build server when its left to choose one.

Benjamin-Knight:
Looking in totally the wrong place, there is nothing wrong with the requirements file, its the testing setup that is incorrect.

We are trying to test against Python 3.8, DBT core does not support this so likely doesn't have the appropriate wheels any longer so Pip fails when trying to install it all now that the versions have creeped higher.

https://docs.getdbt.com/faqs/Core/install-python-compatibility

I'll put in an update to change the workflow to remove those tests.

Benjamin-Knight:
> I'm not opposed to that either. There is a scheduled job that runs the tests on Sundays to check if any fabric merges breaks this adapter. I could make it nightly though if that's better.

The issue is that whilst we may detect the issue unless we fix it quickly anyone installing the adapter, or CI/CD process that runs it is going to pull down the updated fabric version and have code that could potentially not work.

ataft:
I agree with pinning to an exact dbt-fabric version as it appears that dbt-fabric 1.9.5 also introduced an issue, something with access tokens.

I opened a similar PR (https://github.com/dbt-msft/dbt-sqlserver/pull/596), but it looks like this one adds some extra goodies, so I'd be happy to close mine in favor of this.

Any ETA on landing?

axellpadilla:
just to add that this should also close #600

cody-scott:
One other change here might also be to remove the version pins for 3.7 and 3.8 in this file.
https://github.com/Benjamin-Knight/dbt-sqlserver/blob/dbt-1.9/.github/workflows/publish-docker.yml#L15C26-L15C40

cody-scott:
@ataft and @axellpadilla i think closing this one first would be ideal since @Benjamin-Knight has put a ton of great work to bringing it up to 1.9. Once thats through we can check the other PRs off.

The drift between Fabric + SQL Server is problematic, but was inevitable :(

Benjamin-Knight:
> One other change here might also be to remove the version pins for 3.7 and 3.8 in this file. https://github.com/Benjamin-Knight/dbt-sqlserver/blob/dbt-1.9/.github/workflows/publish-docker.yml#L15C26-L15C40

Okay done so.

As this pull request keeps expecting the unit tests for 3.8 to pass and its also trying to remove the unit tests I split out the removal of 3.8 into another pull request so that it can be merged in separately.

https://github.com/dbt-msft/dbt-sqlserver/pull/608

If you are able to force the workflow to pass then I'll close that off otherwise if you merge that first it should let us progress the merging of this request.

Benjamin-Knight:
@cody-scott Should I have done anything on this branch to get a new tag and release onto the main branch or is that something you do after the merge?

PR #604, opened 2025-03-27
Update SQL server relations to not require an alias

### LINKED ISSUE #595: --Empty flag only useful in the simplest projects
Right now the SQL server adapter is inheriting the behaviour of the _render_limited_alias from the DBT core adapter

https://github.com/dbt-labs/dbt-adapters/blob/48c110f547b4e6eea294a179cb562fef5bd29ba4/dbt-adapters/src/dbt/adapters/base/relation.py#L60

This means that all refs get an alias output when running with the empty flag. This is fine if you use no aliases at all in your project, it would seem to me that only the simplest of projects could get away with this requirement. Is anyone actually able to do this successfully?

I would suggest that we change the require_alias flag to false for SQL server and make the provision of aliases down to the project owner. This means that a project that has a requirement for an alias anywhere in their codebase can use the --Empty flag by ensuring aliases are used consistently throughout the project.

Ideally this would be a configurable flag that can be set by the user of the adapter I cannot see how you pass project level flags into BaseRelation and derived classes. If someone knows of a way to do that then I can add a pull request that adds this custom flag and uses that instead.

Benjamin-Knight:
Closes #595

Any project that uses the empty flag and expect the adapter to set aliases would find it no longer works.


axellpadilla:
@Benjamin-Knight  Please merge if ready, they reverted to true this for some reason for v1.10 https://github.com/microsoft/dbt-fabric/commit/4aca77dd5051e5a14dfa785ab5fabeee45cd6dc8
I added a workaround but that works only for empty and samples, and not for microbatching, this should be enough to avoid moving that important detail, microbatch can still be used on pure WITH statements

Benjamin-Knight:
I'll double check this statement they have in the pull request for 1.10. 

Updated the require_alias attribute in FabricRelation to True, aligning with dbt-core 1.10.0 requirements.

I do not remember seeing that requirement but will ensure that is the case.

axellpadilla:
> I'll double check this statement they have in the pull request for 1.10.
> 
> Updated the require_alias attribute in FabricRelation to True, aligning with dbt-core 1.10.0 requirements.
> 
> I do not remember seeing that requirement but will ensure that is the case.

Please let me know if we can include the workaround on v1.10 so feature is usable without having to analyze the implications of the require_alias change, we could find another solution or wait for the bug to be resolved upstream somehow https://github.com/dbt-labs/dbt-adapters/issues/199.

Benjamin-Knight:
I can't see any reason for that statement in the fabric code, the flag you've setup allows for it to be swapped depending on how the user wants the adapter to work. Its important because right now the adapter requires that you either never use an alias or always use an alias to use these features. 

That's fine but right now without the toggle its enforces that you never use an alias, and sometimes that isn't possible when your projects get more complex.

The only question is whether we should have the default be the current behaviour, which is that it does generate an alias and the flag should opt into the new behaviour so that this isn't a breaking change.

axellpadilla:
> I can't see any reason for that statement in the fabric code, the flag you've setup allows for it to be swapped depending on how the user wants the adapter to work. Its important because right now the adapter requires that you either never use an alias or always use an alias to use these features.
> 
> That's fine but right now without the toggle its enforces that you never use an alias, and sometimes that isn't possible when your projects get more complex.
> 
> The only question is whether we should have the default be the current behaviour, which is that it does generate an alias and the flag should opt into the new behaviour so that this isn't a breaking change.

Lets start without changing that so we can test on production and follow up the upstream issue, they could introduce changes or features that depends on having that setup so opt-in is safer right now

Issue #446, opened 2023-09-04
Add column does not work for nvarchar


ericmuijs:
When using the function sqlserver__alter_relation_add_remove_columns for a new field with nvarchar type, a nvarchar(1) is created. 
For any nvarchar field with lengths greater than 1 the resulting merge/insert will fail.

MLSteinw:
I have also come across issues relating to nvarchar and the length of the field not being captured when using the get_column_in_relation function

ericmuijsvanoord:
@cody-scott  : this is still the case for the newest version. I can see if I can work on a fix.

PR #605, opened 2025-04-07
sqlserver_column.py: Handle string dtype of nvarchar

Cogito:
nvarchar is not considered a string in the base adapter, nor in the fabric adapter. This leads to issues when dealing with nvarchar columns, such as in https://github.com/dbt-msft/dbt-sqlserver-utils/blob/master/macros/sql/union.sql

In this commit we add nvarchar to the list of string dtypes, return the correct string_size for nvarchar (which is doubled by default), and use the correct dtype in string_type.

I believe this mishandling of nvarchar is the root cause of #446 

I ran into this issue when unioning multiple relations which used nvarchar. The generated code had no string length specified for the nvarchar fields, which resulted in the default string length of 30 being used.

### DIFF
--- a/dbt/adapters/sqlserver/sqlserver_column.py
+++ b/dbt/adapters/sqlserver/sqlserver_column.py
@@ -20,3 +20,31 @@ class SQLServerColumn(FabricColumn):
             "serial8",
             "int",
         ]
+
+    def is_string(self) -> bool:
+        return self.dtype.lower() in [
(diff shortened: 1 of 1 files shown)


Cogito:
I'm not sure if how I handle `string_type` is best, as it is a class method in the base class, but I think we want it to be instance specific as that is how the dtypes are handled.

PR #606, opened 2025-04-07
sqlserver_column.py: Handle string dtype of nvarchar

Cogito:
nvarchar is not considered a string in the base adapter, nor in the fabric adapter. This leads to issues when dealing with nvarchar columns, such as in https://github.com/dbt-msft/dbt-sqlserver-utils/blob/master/macros/sql/union.sql

In this commit we add nvarchar to the list of string dtypes, return the correct string_size for nvarchar (which is doubled by default), and use the correct dtype in string_type.

I believe this mishandling of nvarchar is the root cause of https://github.com/dbt-msft/dbt-sqlserver/issues/446

I ran into this issue when unioning multiple relations which used nvarchar. The generated code had no string length specified for the nvarchar fields, which resulted in the default string length of 30 being used.

I'm not sure if how I handle string_type is best, as it is a class method in the base class, but I think we want it to be instance specific as that is how the dtypes are handled.

Additionally, I'm not sure why char_size is twice the string length for nvarchar but I noticed it in my testing which is why I halve it in string_size(). I do understand that nvarchar uses twice the bytes as varchar, so suspect it is related, but I don't know if we would want to fix char_size to be the actual number of characters or just fix it in string_size.

### DIFF
--- a/dbt/adapters/sqlserver/sqlserver_column.py
+++ b/dbt/adapters/sqlserver/sqlserver_column.py
@@ -20,3 +20,31 @@ class SQLServerColumn(FabricColumn):
             "serial8",
             "int",
         ]
+
+    def is_string(self) -> bool:
+        return self.dtype.lower() in [
+            "text",
+            "character varying",
+            "character",
+            "varchar",
+            "nvarchar",
+        ]
+
+    def string_size(self) -> int:
+        if not self.is_string():
+            raise DbtRuntimeError("Called string_size() on non-string field!")
+
+        if self.dtype == "text" or self.char_size is None:
+            # char_size should never be None. Handle it reasonably just in case
+            return 256
+        elif self.dtype.lower() == "nvarchar":
+            # char_size is doubled for nvarchar
+            return int(self.char_size // 2)
+        else:
+            return int(self.char_size)
+
+    def string_type(self, size: int) -> str:
+        if self.dtype:
+            return f"{self.dtype}({size if size > 0 else '8000'})"
+        else:
+            return f"varchar({size if size > 0 else '8000'})"
--- a/dbt/adapters/sqlserver/sqlserver_column.py
+++ b/dbt/adapters/sqlserver/sqlserver_column.py
@@ -1,4 +1,5 @@
 from dbt.adapters.fabric import FabricColumn
+from dbt_common.exceptions import DbtRuntimeError
 
 
 class SQLServerColumn(FabricColumn):
--- a/dbt/include/sqlserver/macros/adapter/columns.sql
+++ b/dbt/include/sqlserver/macros/adapter/columns.sql
@@ -48,3 +48,36 @@
     {% do run_query(rename_column) %}
 
 {% endmacro %}
+
+{% macro sqlserver__get_columns_in_relation(relation) -%}
+    {% set query_label = apply_label() %}
+    {% call statement('get_columns_in_relation', fetch_result=True) %}
+        {{ get_use_database_sql(relation.database) }}
+        with mapping as (
+            select
+                row_number() over (partition by object_name(c.object_id) order by c.column_id) as ordinal_position,
+                c.name collate database_default as column_name,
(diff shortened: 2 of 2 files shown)


Cogito:
apologies for the noise, I renamed the branch which closed the first pull request

Benjamin-Knight:
> Additionally, I'm not sure why char_size is twice the string length for nvarchar but I noticed it in my testing which is why I halve it in string_size(). I do understand that nvarchar uses twice the bytes as varchar, so suspect it is related, but I don't know if we would want to fix char_size to be the actual number of characters or just fix it in string_size.

nvarchar uses twice the number byte pairs to store a characters as it has the extended character set, hence it only goes to nvarchar(4000) compared to varchar(8000) even though they take the same amount of storage space.

Cogito:
My reading of the existing code is that char_size should be the declared maximum number of characters allowed, not the number of bytes allocated.

Anywhere it is used it is assumed to be the number of characters. To date I doubt many are using it at all, as nvarchar is not currently a string type so the width of any columns being created is being set to the default (which is 30 I think everywhere?)

As per https://stackoverflow.com/questions/37187654/alternative-to-sys-columns-max-length we could potentially use INFORMATION_SCHEMA.COLUMNS instead of sys.columns in https://github.com/microsoft/dbt-fabric/blob/d3b8665f4e116d0fcbd14e544071f5d3cf7eb15e/dbt/include/fabric/macros/adapters/columns.sql#L9

This would probably require us maintaining a fork of that macro in https://github.com/dbt-msft/dbt-sqlserver/blob/master/dbt/include/sqlserver/macros/adapter/columns.sql unless we can get it fixed in the fabric adapter. Fabric doesn't have nvarchar datatypes so not sure if it makes sense for them to fix it there or not.

Benjamin-Knight:
A SQL server specific version of the macro that gets the data seems the safest method to me, as you say it will not be implemented at the Fabric level.

That should mean that the only code change would need to be the update to is_string to handle nchar and nvarchar.

We will want a variation of https://github.com/microsoft/dbt-fabric/blob/main/tests/functional/adapter/test_column_types.py as well to support the new nvarchar and nchar handling.

Cogito:
bumping this, what is needed to get it merged? @Benjamin-Knight do you know?

axellpadilla:
Thanks for the patch. I re-checked both official `master` and this branch and adding this note for future use.

Official `master` already has the SQL metadata part of the fix in `sqlserver__get_columns_in_relation`: it reads from `sys.columns` / `sys.types` and divides `max_length` by 2 for `nchar`, `nvarchar`, and `sysname`, which is correct because SQL Server reports those lengths in bytes.

The remaining gap on official `master` is in `SQLServerColumn`: `is_string()` only includes `varchar` and `char`, so `nvarchar` / `nchar` can still fall through to `return self.dtype`. In the add-column path, that can emit bare `nvarchar`, which SQL Server treats as `nvarchar(1)`.

This branch fixes that core issue by adding `nvarchar` and `nchar` to `is_string()` and rendering string types through an instance-aware method, so `nvarchar(100)` is emitted as `nvarchar(100)` instead of bare `nvarchar`.

For the final PR, I suggest keeping the scope narrow:

* keep the current `master` metadata query behavior;
* keep the instance-aware rendering for `nvarchar` / `nchar`;
* include `nvarchar` and `nchar` in `is_string()`;

We need more fixes for different types that I'm using on my personal patched to reach feature parity, but this means tables get alter tables and could even try to alter enourmous tables, we should gate type expansion for big tables @Benjamin-Knight , that is what I did to avoid problems, big tables change of types should be handled by the admins instead of running posible long locking or dead locks alters, but this is another issue.

Cogito:
Hi @axellpadilla, do we need updates made to this branch? What you've written makes sense but not sure if changes needed here.

axellpadilla:
@Cogito an expanding this  and adding a feature flag for the new behavior, related, #701, this is superseeded by a new pr

PR #608, opened 2025-04-12
Update testing matrix

Benjamin-Knight:
Update the unit and integration tests to no longer test against Python 3.8 as DBT does not support this version.


Benjamin-Knight:
Close as completed via #598

PR #610, opened 2025-05-03
Fix: Handle view materialization alterations

axellpadilla:
This pr addresses the following:

- Modifies the create view macro to use **create or alter** view when the view already exists, preventing errors during re-materialization.
- Updates the view materialization logic to correctly avoid renaming if no table exists.
- Adds test cases to verify that materializing an existing view updates its definition correctly, including column changes, previous grants and avoids table changes if view fails.
- Added grants migration and tests
- Added no-op when view needs no changes to avoid locks on schemas


axellpadilla:
Hi @cody-scott , did you have any time to review this? Maybe you could authorize some extra hands to make reviews or even maintainers, thing is, on sql server a drop/create view are 2 steps and plenty of system tables modifications, while alter is 1 step and less overhead, this should be a high impact change for big dbt projects (it is for my use cases).

dbt/include/sqlserver/macros/relations/views/create.sql
Benjamin-Knight, line 1: Could we replace the loading of the cached relation and the branching with a simple CREATE OR ALTER VIEW statement? Supported since SQL server 2016 and the only version older than that which is supported is SQL 2014 with some very long term extended support contract in place so it seems safe to assume everyone has this syntax available.

dbt/include/sqlserver/macros/materializations/models/view/view.sql
Benjamin-Knight, line 1: You've changed this and are no longer using the intermediates that this macro creates and have swapped the build order.

Because you are not renaming the old relation and then building the new one, if that build fails its inconsistent.

That leaves the risk of if the view is not valid that it will clean up the existing relation and then not create the new one corrupting the deployment, I think this would work for the view->view that you have the test for. But if we are going from a table->view I think this could break.

Benjamin-Knight (dismissed):
Based on the potential issue with the view.sql changes I think this change is missing 3 test paths, 1 is optional, that the base dbt tests are not covering. 

Going from a table -> view needs checking.
Ensuring that if the new view definition is incorrect that we do not break the deployment.

The optional path would be that the request has 'non-dbt grants deletion' as part of the reason to make this change. You could add in a test to ensure that this path preserves these, partly as a regression test but also because of the all the intermediate objects in the macro to ensure that grants persist across that dance.,

dbt/include/sqlserver/macros/materializations/models/view/view.sql
axellpadilla: you were right thanks, adding a test to test rollback

axellpadilla:
@Benjamin-Knight added the no-op please check if a good solution and merge or change when ready

Benjamin-Knight:
That new grant logic, it takes grants from tables and applies to to views, have you tested it with grants that would only apply for tables like an insert grant? I think SQL will complain if you try and grant something like that to a view.

axellpadilla:
> That new grant logic, it takes grants from tables and applies to to views, have you tested it with grants that would only apply for tables like an insert grant? I think SQL will complain if you try and grant something like that to a view.

SQL Server natively allows you to grant INSERT, UPDATE, and DELETE permissions on views, just as you would on a table. The GRANT statement itself will execute perfectly fine. The action of actually using it probably not if the view isn't writable.

axellpadilla:
@Benjamin-Knight hardened the concept, it only passes waranted compatible grants, should be ready to merge after v1.9 branched

joshmarkovic:
This is an excellent change! Let me know if I can help get this over the finish line in any way

axellpadilla:
> This is an excellent change! Let me know if I can help get this over the finish line in any way

Yes please review it, if you think is ready let us know, feedback is appreciated, we already preparing this for v1.10

PR #617, opened 2025-07-14
Support new encryption options

LuisMunozVillarreal:
As per the documentation, this change adds support for all the new encryption options

https://learn.microsoft.com/en-us/sql/relational-databases/security/networking/tds-8?view=sql-server-ver17#additional-changes-to-connection-string-encryption-properties  # noqa: E501
EncryptType = Literal[
    "true",
    "yes",
    "mandatory",
    "strict",
    "optional",
    "false",
    "no",
    True,
    False,

### DIFF
--- a/dbt/adapters/sqlserver/sqlserver_connections.py
+++ b/dbt/adapters/sqlserver/sqlserver_connections.py
@@ -103,7 +103,10 @@ class SQLServerConnectionManager(FabricConnectionManager):
         assert credentials.encrypt is not None
         assert credentials.trust_cert is not None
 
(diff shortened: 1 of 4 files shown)


LuisMunozVillarreal:
This change passed tests on my fork while using this repo's docker images:
- Unit tests: https://github.com/LuisMunozVillarreal/dbt-sqlserver/actions/runs/16273179409
- Integration: https://github.com/LuisMunozVillarreal/dbt-sqlserver/actions/runs/16273179406

Last commit fails because I reverted the workflows to use the current repo images and my fork doesn't have the necessary images

PR #623, opened 2025-09-29
Bump dbt-fabric to v1.9.6 and fix SQL Server unit-test

axellpadilla:
Bump dbt-fabric to v1.9.6.
Port Fabric PR #315: add sqlserver-specific unit-test materialization and nested-CTE detection (sqlserver__unit_test_create_table_as, unit materialization). Use an intermediate view + SELECT INTO / CREATE+INSERT to avoid EXEC-wrapped CTAS and ensure labels are applied inside valid T-SQL. Prepare for Python 3.13 compatibility (support pyodbc >= 5.2 required by fabric 1.9.6).

Issues linked: 
#612
#618

PR #628, opened 2026-01-05
Remove Fabric Dependency

### LINKED ISSUE #612: Support for Python 3.13
Python 3.13 was released on October 2024. Would be nice to have `dbt-sqlserver` available for it too.

### LINKED ISSUE #614: Warning! of breaking change on dbt-fabric
Around or before this change on upstream repo, was caused a breaking change: 
https://github.com/microsoft/dbt-fabric/pull/270

While this makes both adapters behave like the default dbt docs, and is good, it was indeed a breaking change, some database now will have duplicated tables and views, and even could have errors.

This is the default:
```jinja
{% macro generate_schema_name(custom_schema_name, node) -%}

    {%- set default_schema = target.schema -%}
    {%- if custom_schema_name is none -%}

        {{ default_schema }}

    {%- else -%}

        {{ default_schema }}_{{ custom_schema_name | trim }}

    {%- endif -%}

{%- endmacro %}
```

This is what was probably removed in x change:
```jinja
{% macro fabric__generate_schema_name(custom_schema_name, node) -%}

    {%- set default_schema = target.schema -%}
    {%- if custom_schema_name is none -%}

        {{ default_schema }}

    {%- else -%}

        {{ custom_schema_name | trim }}

    {%- endif -%}

{%- endmacro %}
```

This should be investigated further and documented so old version users know they need to have that macro overrided to preserve previous schema naming.

Benjamin-Knight:
This pull request is a big one that aims to remove the dependency this adapter has on the dbt-fabric adapter.

As Fabric drifts further from SQL server it will become increasingly difficult to keep Fabric as a base, its also still on version 1.9.

I've also looked at the work @axellpadilla has been doing recently and added in the CI work to support python 3.13 with a view to bumping this adapter up to version 1.11 to match the current DBT core release.

@cody-scott As the most recently active maintainer are you able to help progress this issue or hand additional rights to myself or @axellpadilla to see if we can start closing off some of the issues on this adapter?

In terms of the pull request as its a big one the more eyes on it the better, I've ported over some of the functional tests from Fabric that were not implemented in this adapter to try and keep coverage up and all current functional tests we had are passing. I think we may still want to do a beta release however.


seb-afk:
@Benjamin-Knight I am just an extremely happy user of this adapter but if there is something I could potentially help with please let me know and I will check with my employer if I can carve out some time to work, do tests, etc on this adapter.

cody-scott:
@Benjamin-Knight can you resolve the conflicts on this merge. I merged the other request to get the 3.13 tests to pass but there is some conflicts.

Benjamin-Knight:
> @Benjamin-Knight can you resolve the conflicts on this merge. I merged the other request to get the 3.13 tests to pass but there is some conflicts.

Done

Benjamin-Knight:
The previous update by @axellpadilla did some test changes, I also did those but moved the location of the tests so it duplicated a macro, I've fixed that up but there is a new materialization for the unit created in unit_test_table.

I do not know the purpose of this but this new materialization is causing some test failures, removing the materialization causes tests to all pass again.

axellpadilla:
@Benjamin-Knight As I remember, that test materialization was needed because some compatibility issues with upstream, moving out from fabric could mean its not now.

Benjamin-Knight:
> @Benjamin-Knight As I remember, that test materialization was needed because some compatibility issues with upstream, moving out from fabric could mean its not now.

Ok removed it, all the tests pass my end.

Benjamin-Knight:
Must admit I'm a bit confused as to why its still trying to do 3.8 unit tests, they are in the workflow any longer. Assume its baked into some sort of pull guard rail.

https://github.com/dbt-msft/dbt-sqlserver/blob/master/.github/workflows/unit-tests.yml

axellpadilla:
You are right, but reviewing it made realize we need to update setup.py

setup.py
axellpadilla:
```suggestion
```

axellpadilla (dismissed):
Remove python <=3.8 from setup.py

setup.py
Benjamin-Knight: Done

axellpadilla:
This replaces #623 right?
Closes:
https://github.com/dbt-msft/dbt-sqlserver/issues/612
https://github.com/dbt-msft/dbt-sqlserver/issues/618

axellpadilla:
Also, for release we should document this change because this will default to dbt docs, and its this way without warning from last version:
#614

Benjamin-Knight:
> This replaces #623 right? Closes: #612 #618

It doesn't implement 1.10 yet so #618 will still be active, I didn't want to try and remove fabric and bump the version all at once.

Benjamin-Knight:
> Also, for release we should document this change because this will default to dbt docs, and its this way without warning from last version: #614

We could also use a behaviour flag to restore the old functionality, if we felt it was that much of a breaking change we could leave the old behaviour in and gate the new behaviour behind a flag to let people opt into it until we deprecate in a later release, this is how DBT core handles this.

axellpadilla:
I wasn't able to exactly get where this change happened (version), it is indeed breaking (even duplicates dwh data because the different schema name), but I support the idea of behavior change, if we can find the exact part and version that decides if by default or not on this case, on my PRD I needed to leave it by default because moving a whole dwh schema names to new ones is simply not worth

Benjamin-Knight:
> I wasn't able to exactly get where this change happened (version), it is indeed breaking (even duplicates dwh data because the different schema name), but I support the idea of behavior change, if we can find the exact part and version that decides if by default or not on this case, on my PRD I needed to leave it by default because moving a whole dwh schema names to new ones is simply not worth

It was here.

https://github.com/microsoft/dbt-fabric/pull/251

Its not listed as a change so I suspect it was not intended.

axellpadilla:
> > I wasn't able to exactly get where this change happened (version), it is indeed breaking (even duplicates dwh data because the different schema name), but I support the idea of behavior change, if we can find the exact part and version that decides if by default or not on this case, on my PRD I needed to leave it by default because moving a whole dwh schema names to new ones is simply not worth
> 
> It was here.
> 
> [microsoft/dbt-fabric#251](https://github.com/microsoft/dbt-fabric/pull/251)
> 
> Its not listed as a change so I suspect it was not intended.

Ouch, fabric 1.8.8, that means we have on our side the breaking change on any version of sql server that updated fabric after https://github.com/dbt-msft/dbt-sqlserver/commit/a204adb9597c744b4f71636c42b4977d8d878613
2 years ago where fabric was opened to "dbt-fabric>=1.8.0,<1.9.0", I think most people would have already handled or impacted by this on 2025, should dbt default be the new standard and just documented on the release?.

mopthe:
@axellpadilla I agree, defaulting to the new standard makes sense.

Back then I think we handled it like this as a quick-fix:

```yml
packages:
  - package: microsoft/dbt-fabric
    version: ">=1.8.0,<1.9.0,!=1.8.8"
```

This way we stay in the 1.8.x range but explicitly skip the problematic 1.8.8 version.

axellpadilla:
> @axellpadilla I agree, defaulting to the new standard makes sense.
> 
> Back then I think we handled it like this as a quick-fix:
> 
> ```yaml
> packages:
>   - package: microsoft/dbt-fabric
>     version: ">=1.8.0,<1.9.0,!=1.8.8"
> ```
> 
> This way we stay in the 1.8.x range but explicitly skip the problematic 1.8.8 version.

So it was a single version problem? I never looked back, just patched the whole macro on project, if that was a single version problem then this release could introduce that breaking change if we don't add a work-around, default right now is https://github.com/dbt-labs/dbt-adapters/blob/05eaf6e2c4c32620b6dcff39e372154a9b6125f5/dbt-adapters/src/dbt/include/global_project/macros/get_custom_name/get_custom_schema.sql#L21

so, final idea is to call default macro if flag is set but old if not set:
```jinja
{% macro sqlserver__generate_schema_name(custom_schema_name, node) -%}
    {%- if var('dbt_sqlserver_use_default_schema_concat', false) -%}
        {{ default__generate_schema_name(custom_schema_name, node) }}
    {%- else -%}
        {%- set default_schema = target.schema -%}
        {%- if custom_schema_name is none -%}
            {{ default_schema }}
        {%- else -%}
            {{ custom_schema_name | trim }}
        {%- endif -%}
    {%- endif -%}
{%- endmacro %}
```

Readme:
````md
## Configuration

### Flags

- `dbt_sqlserver_use_default_schema_concat`: (default: `false`) When `true`, uses dbt-core's standard schema concatenation. When `false` (default), uses legacy adapter behavior. Provides backward compatibility.

  **Example usage in `dbt_project.yml`:**
  ```yaml
  flags:
    dbt_sqlserver_use_default_schema_concat: true  # Enable standard schema concatenation
  ```

Alternatively, users can override the `sqlserver__generate_schema_name` macro in their project to customize schema generation if they prefer not to use the flag or to avoid problem on deprecation of this feature.
  
````

impl.py
```python
from dbt_common.behavior_flags import BehaviorFlag

DBT_SQLSERVER_USE_DEFAULT_SCHEMA_CONCAT = BehaviorFlag(
    name="dbt_sqlserver_use_default_schema_concat",
    default=False,
    description=(
        "When True, uses default dbt schema concatenation. When False (default), uses legacy adapter behavior, "
        "Combining `target.schema` and `custom_schema_name` if provided."
        "For permanent usage of dbt, override the sqlserver__generate_schema_name macro instead."
    ),
)

class SQLServerAdapter(FabricAdapter):
    # ... other code ...

    @property
    def _behavior_flags(self) -> List[BehaviorFlag]:
        return [DBT_SQLSERVER_USE_DEFAULT_SCHEMA_CONCAT]

    # ... rest of the class ...
```

mopthe:
Yes, from what I recall, this only affected a single release (1.8.8) of dbt-core (received via Fabric, so that’s probably dbt-core we pinned). We’ve been using our own macro for a long time now, so I don’t remember all the details anymore.

In the following release, dbt Core reset the constraints in “Move upper bounds back to 2.0” (dbt-labs/dbt-core#11024).

The flag-based approach for backward compatibility seems reasonable, though if this really was isolated to 1.8.8, preserving the legacy behavior by default may not be necessary.

Just wanted to highlight that, you’re the experts.

hslira:
+1

Benjamin-Knight:
> Yes, from what I recall, this only affected a single release (1.8.8) of dbt-core (received via Fabric, so that’s probably dbt-core we pinned). We’ve been using our own macro for a long time now, so I don’t remember all the details anymore.
> 
> In the following release, dbt Core reset the constraints in “Move upper bounds back to 2.0” ([dbt-labs/dbt-core#11024](https://github.com/dbt-labs/dbt-core/pull/11024)).
> 
> The flag-based approach for backward compatibility seems reasonable, though if this really was isolated to 1.8.8, preserving the legacy behavior by default may not be necessary.
> 
> Just wanted to highlight that, you’re the experts.

Not sure I'd call myself an expert, I've implemented the flag based approach, it defaults to preserving legacy behaviour and you opt into the dbt-core expected behaviour as that would be the intent of a behaviour flag, at some point in the future we would swap this round so you would opt into the old behaviour instead.

It may be that we do this for the full release and its only this way around for release candidates but we have that option now.

Benjamin-Knight:
I would also note https://github.com/dbt-msft/dbt-sqlserver/pull/633.

As its written by an actual DBT maintainer it may end up superseding this, the conversion to ADBC is required for DBT Fusion so if it moves us towards that then that would be excellent news, although there are downsides to the ADBC move as noted in the pull request.

mopthe:
Hi @Benjamin-Knight  @cody-scott  @axellpadilla ,

We're running dbt-sqlserver 1.9.0 in production (SQL Server, not Fabric) with a large-scale dbt projec, 2000+ models across multiple orchestrated flows. We're very interested in seeing this PR land.

Regarding #633  (ADBC port): it looks promising long-term, but it's a blocker for us until it supports Active Directory authentication. So from our perspective, this PR is the more actionable path forward right now.

We'd be happy to test a release candidate / beta build against our production workloads if that would help build confidence for a merge. We can validate incremental runs, full-refresh scenarios, and snapshot behavior at scale.

The behavior flag approach for generate_schema_name looks reasonable to us, we already override generate_schema_name in our project so the breaking change wouldn't affect us directly, but defaulting to legacy behavior seems like the right call for a safe rollout.

Is there anything specific blocking the final approval? Happy to help where we can.

joshmarkovic:
> Is there anything specific blocking the final approval? Happy to help where we can.

Echoing this sentiment as well. I am willing to test out a beta/release candidate on a large deployment and would love to get more involved once this lands.

Benjamin-Knight:
@mopthe @joshmarkovic 
I do not have the permissions to progress this, we are reliant currently on @cody-scott having the time to sign this off and get it to a release candidate.
I also have a similar large sized project run internally, I'm actually already using this branch (actually a 1.10 release I've prepared off the back of it) in our development environment as a smoke test and its not burnt down yet.

axellpadilla:
Hi @Benjamin-Knight Already running on PRD also, but needed two small patches that I sent on a PR to your branch, thanks ahead.

Benjamin-Knight:
@cody-scott Are you able to update the branch rules so that the 3.8 test is no longer required, and add 3.12 and 3.13. After that the tests are all green so I think this is ready to go to a release candidate, currently planned to be 1.9.1rc1

axellpadilla (approved):
💯  @Benjamin-Knight As you have previous experience on releases with this project will wait for you to proceed with the release candidate.

Benjamin-Knight:
Released https://pypi.org/project/dbt-sqlserver/#history

Anyone interested in testing do so and log all issues and we should hopefully be able to resolve them and get this to a proper release so we can move onto all the other open issues and bringing us up to DBT 1.11 as version 1.10 is almost end of support.

joshmarkovic:
Amazing! It worked for my codebase (~900 models). The only adjustment I would make is to add `pyodbc` to setup.py (I believe it was removed with `dbt-fabric`). Let me know if you want me to make that change.

Benjamin-Knight:
> Amazing! It worked for my codebase (~900 models). The only adjustment I would make is to add `pyodbc` to setup.py (I believe it was removed with `dbt-fabric`). Let me know if you want me to make that change.

That makes sense, previously we referenced fabric so were getting pyodbc>=5.2.0 via transitive dependencies. We need an issue for it at least, you can do the pull request as well if you want, I know axel was looking at using or supporting different underlying drivers so not sure how we want to handle having a hard dependency in setup so having the issue will be a good spot to discuss that.

PR #630, opened 2026-01-21
fix: enable multikey snapshots

### LINKED ISSUE #615: Snapshots aren't correctly handling multiple unique_key values
# Issue

When [configuring a snapshot with multiple](https://docs.getdbt.com/reference/resource-configs/unique_key#use-multiple-unique-keys) `unique_key` values the adapter does not correctly handle the columns created by the `get_or_create_relation` macro. This results in the 2nd execution adding `dbt_unique_key_1`, `dbt_unique_key_2`, etc. to the snapshot table. On a 3rd execution the `get_or_create_relation` macro produces invalid SQL by using `select *` in addition to aliasing the unique keys again as `dbt_unique_key_1`, `dbt_unique_key_2`, etc.

# Root Cause Analysis

I've tracked down the cause to be this section of code, which is working on the assumption `unique_key` is a single value. As such, the length of `missing_columns` is greater than 0 and the `create_columns` macro is executed.
https://github.com/dbt-msft/dbt-sqlserver/blob/bcf4ac910528c9bc681df5fd487818a19094838a/dbt/include/sqlserver/macros/materializations/snapshot/snapshot.sql#L58-L74

# Suggested Solution

I suggest updating that section similar to how the snapshot materialisation is [implemented in dbt-fabric](https://github.com/microsoft/dbt-fabric/blob/772d896d366137ff1ed1a364110d9eee67a727e6/dbt/include/fabric/macros/materializations/snapshots/snapshot.sql#L63-L79), by creating a list of columns to ignore and appending each `unique_key` value if it happens to be a list.

# Testing

I've monkey-patched this locally and it appears to work as expected with some very limited testing.
For clarity, I've included a comparison of the file with my patch applied.

Image

---

I'd be happy to also raise this as a PR, once I'm not on the work PC, if it makes things easier.

hoesler:
This PR fixes #615 by adding dynamic column filtering for multiple unique keys

### DIFF
--- a/dbt/include/sqlserver/macros/materializations/snapshot/snapshot.sql
+++ b/dbt/include/sqlserver/macros/materializations/snapshot/snapshot.sql
@@ -55,11 +55,16 @@
       {% do adapter.expand_target_column_types(from_relation=staging_table,
                                                to_relation=target_relation) %}
 
+      {% set remove_columns = ['dbt_change_type', 'DBT_CHANGE_TYPE', 'dbt_unique_key', 'DBT_UNIQUE_KEY'] %}
+      {% if unique_key is sequence and unique_key is not string %}
+        {% for key in strategy.unique_key %}
+          {{ remove_columns.append('dbt_unique_key_' + loop.index|string) or "" }}
+          {{ remove_columns.append('DBT_UNIQUE_KEY_' + loop.index|string) or "" }}
+        {% endfor %}
+      {% endif %}
+
       {% set missing_columns = adapter.get_missing_columns(staging_table, target_relation)
-                                   | rejectattr('name', 'equalto', 'dbt_change_type')
-                                   | rejectattr('name', 'equalto', 'DBT_CHANGE_TYPE')
-                                   | rejectattr('name', 'equalto', 'dbt_unique_key')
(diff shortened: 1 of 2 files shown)


axellpadilla:
Thanks, closing because the current master version and (v1.9.1) of:

dbt/include/sqlserver/macros/materializations/snapshots/snapshot.sql

already contains the multi-key exclusion logic:

{% set remove_columns = ['dbt_change_type', 'DBT_CHANGE_TYPE', 'dbt_unique_key', 'DBT_UNIQUE_KEY'] %}

{% if unique_key | is_list %}
  {% for key in strategy.unique_key %}
    {{ remove_columns.append('dbt_unique_key_' + loop.index|string) }}
    {{ remove_columns.append('DBT_UNIQUE_KEY_' + loop.index|string) }}
  {% endfor %}
{% endif %}

{% set missing_columns = adapter.get_missing_columns(staging_table, target_relation)
  | rejectattr('name', 'in', remove_columns)
  | list %}

{% set source_columns = adapter.get_columns_in_relation(staging_table)
  | rejectattr('name', 'in', remove_columns)
  | list %}

PR #631, opened 2026-02-04
Create packages.yml

BindusekharGorintla:
The packages.yml file in dbt can declare external dbt projects (called packages) that your project depends on, allowing you to reuse models, macros, and tests from other developers. This file is placed at the root level of your dbt project, alongside your dbt_project.yml file

### DIFF
--- a/packages.yml
+++ b/packages.yml
@@ -1,0 +1,9 @@
+packages:
+  - package: dbt-labs/dbt_utils
+    version: 1.1.1
+  - package: dbt-labs/codegen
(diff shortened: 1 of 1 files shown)


mopthe:
`packages.yml` is not something that belongs in this repository.

In dbt, `packages.yml` is a project‑level configuration file that each *individual dbt project* defines for itself, alongside its own `dbt_project.yml`. It should not be added to adapter source code, because adapters are not dbt projects and do not declare dependencies on behalf of users.

PR #643, opened 2026-05-14
chore: Add weekly stale cleanup workflow

axellpadilla:
helps with stale requests and review of remaining issues

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,38 @@
+---
+name: Close stale issues and PRs
+on:  # yamllint disable-line rule:truthy
+  workflow_dispatch:
+  schedule:
+    - cron: 30 1 * * 1  # Every Monday at 1:30 AM UTC
+permissions:
+  issues: write
+  pull-requests: write
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v10
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          days-before-issue-stale: 365
+          days-before-pr-stale: 365
+          days-before-issue-close: 30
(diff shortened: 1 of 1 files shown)


Benjamin-Knight:
A good idea, but maybe we should do a manual pass over all the current open issues before we activate this?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,38 @@
...
+          stale-pr-message: >-
+            This pull request is stale because it has had no activity for 365 days.
+            Please confirm whether it is still relevant. If there is no response,
+            it will be closed in 30 days.
+          close-issue-message: >-
+            Closing this issue due to inactivity. Please reopen or create a new issue
+            if this is still needed.
+          close-pr-message: >-
+            Closing this pull request due to inactivity. Please reopen or create a
+            new
```
Benjamin-Knight: Unusual line break?

.github/workflows/stale.yml
Benjamin-Knight, line 1: Do we want to exempt issues with a draft PR request? We would want to close off the PR before we marked an issue as stale?

Does the actions/stale action create the stale label because that isn't a label in this repo right now.

Benjamin-Knight (commented):
I have not setup git hub actions before so I wouldn't rely on my knowledge of these but I noted a few things.

.github/workflows/stale.yml
axellpadilla: you are right, it was the formatter

.github/workflows/stale.yml
axellpadilla: I'm using 365 days just to close old ones that are probably not even needed or already solved, we shouldn't worry about the first issue.
Mmm second one, yes, it is supposed to work as a plug and play

PR #645, opened 2026-05-14
Modernize packaging: migrate setup.py to PEP 621 pyproject.toml

joshmarkovic:
- Replace setup.py with pyproject.toml (setuptools backend), reading version dynamically from dbt/adapters/sqlserver/__version__.py.
- Restore pyodbc>=4.0.35,<5.2.0 as a direct dependency. It was previously pulled in transitively via dbt-fabric; that dep was dropped in 1.9.1rc1 but the runtime `import pyodbc` was not, leaving fresh installs broken.
- Move azure-identity to an `azure` extra. The code already guards azure imports with try/except ModuleNotFoundError, so this makes the optionality first-class: pip install dbt-sqlserver[azure].
- Move dev tooling to a `dev` extra; retire dev_requirements.txt. Use pip install -e .[dev].
- Drop MANIFEST.in in favor of [tool.setuptools.package-data].
- Ship dbt/adapters/sqlserver/py.typed.
- requires-python = ">=3.10"; drop the Python 3.9 classifier (3.9 reached EOL in October 2025).
- Update unit/integration/release workflows + Makefile accordingly. The release workflow now uses `python -m build` and an inline tag/version match instead of `python setup.py sdist bdist_wheel` and the old VerifyVersionCommand.


.github/workflows/release-version.yml
```diff
@@ -15,13 +15,29 @@ jobs:
...
+          python-version: '3.10'
 
-      - name: Install dependencies
-        run: pip install -r dev_requirements.txt
+      - name: Install build tooling
+        run: pip install build twine
 
-      - name: Verify version match
-        run: python setup.py verify
+      - name: Verify version matches tag
```
Benjamin-Knight, line 23: Can we move this python script to an actual script rather than inline? Right now this python code can't get tested or go through our normal python pre-commit steps.

pyproject.toml
```diff
@@ -0,0 +1,77 @@
...
+    "dbt-adapters>=1.11.0,<2.0",
+    "pyodbc>=4.0.35,<5.2.0",
+]
+dynamic = ["version"]
+
+[project.optional-dependencies]
+azure = [
+    "azure-identity>=1.12.0",
+]
+dev = [
```
Benjamin-Knight, line 45: Do we want people to be able to do dbt-sqlserver[dev] to install? Would it make more sense to put these into dependency-groups so that consumers of the package can't install the dev stuff, people developing could do uv sync --group dev?

axellpadilla: this is the correct structure example:
[project.optional-dependencies]
external = [...]
another = [...]

[dependency-groups]
dev = [...]

axellpadilla:
@joshmarkovic I migrated this to uv, thats why you have conflicts, PEP 621 + 735.
https://docs.astral.sh/uv/concepts/projects/dependencies/#project-dependencies
https://docs.astral.sh/uv/concepts/projects/dependencies/#optional-dependencies
https://docs.astral.sh/uv/concepts/projects/dependencies/#development-dependencies

pyproject.toml
```diff
@@ -0,0 +1,79 @@
...
+    "dbt-adapters>=1.11.0,<2.0",
+    "pyodbc>=4.0.35,<5.2.0",
+]
+dynamic = ["version"]
+
+[project.optional-dependencies]
+azure = [
+    "azure-identity>=1.12.0",
+]
+
```
axellpadilla, line 43: please move pyodbc to optional and pyodbc>=5.2.0, we are going to add mssql-python as optional later too

axellpadilla (approved):

PR #660, opened 2026-05-16
fix: Use uv to create and activate a workspace .venv to avoid permissions errors

axellpadilla:
Final fix to devcontainer behavior, some devcontainers make permissions strictier.

This also allows non-devcontainer easier setup.

### DIFF
--- a/.devcontainer/setup_env.sh
+++ b/.devcontainer/setup_env.sh
@@ -1,14 +1,14 @@
+set -euo pipefail
+
 cp test.env.sample test.env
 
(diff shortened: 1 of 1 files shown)


axellpadilla:
@Benjamin-Knight this should fix all the issues, before this it still wasn't building on one of my machines (but was ok in the first one)

axellpadilla:
@Benjamin-Knight and another one about a missing libodbc.so.2, perfectly fine... fixing it already

PR #661, opened 2026-05-16
fix: devcontainer post-start sequentially 

axellpadilla:
Because uv runs too fast and stops the background odbc install after ending... thanks uv


axellpadilla:
@Benjamin-Knight finally

PR #664, opened 2026-05-19
Bump v1.9.1 release and keeps pyodbc temporary

axellpadilla:
To avoid unwanted and non-announced dependency removals o some cases we needed to keep pyodcb dependency temporary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,26 @@
 # Changelog
 
+### v1.9.1
+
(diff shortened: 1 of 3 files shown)


axellpadilla:
@Benjamin-Knight Please review the final state and create the release if everything ok, I also created the branch release/v1.9 (up to date to this commit) to move from this step and provide any bugfix if needed to be able to merge v1.10 changes without this in the middle

PR #674, opened 2026-05-20
Add query_options model config for SQL Server OPTION clauses (#613).

### LINKED ISSUE #613: Max Recursion config property
We have come across some tables that require a recursion limit of greater then 100, the SQL Server default. Proposing to add a config property to insert a variable max recursion amount.

The current adapter implements the options here
https://github.com/dbt-msft/dbt-sqlserver/blob/master/dbt/include/sqlserver/macros/adapter/metadata.sql

Since the option `maxrecursion` cannot be apart of the create view statement, this should only apply to the create table statements. Based on how this is set now it would either need to be changed for all calls to check if its a table materialization or we have an alternative table version that adds maxrecursion.

My suggestion would be to have it always include the maxrecursion in the option, but have the default be 100 to mimic the sql server default.

apply_label appears in a few spots, so perhaps the simpler option is have `apply_label_for_table` as a starting point. We wouldn't have to change the apply_label signature to require a materialization type argument in that case.

https://github.com/search?q=repo%3Adbt-msft%2Fdbt-sqlserver%20apply_label&type=code

Option
```jinja
{% macro apply_label() %}
    {{ log (config.get('query_tag','dbt-sqlserver'))}}
    {%- set query_label = config.get('query_tag','dbt-sqlserver') -%}
    OPTION (LABEL = '{{query_label}}');
{% endmacro %}

{% macro apply_label_for_table() %}
    {{ log (config.get('query_tag','dbt-sqlserver'))}}
    {%- set query_label = config.get('query_tag','dbt-sqlserver') -%}
    {%- set max_recursion = config.get('max_recursion',100) -%}
    OPTION (
        LABEL = '{{query_label}}', 
        MAXRECURSION = {{ max_recursion }}
    );
{% endmacro %}
```

then change the call here: https://github.com/dbt-msft/dbt-sqlserver/blob/bcf4ac910528c9bc681df5fd487818a19094838a/dbt/include/sqlserver/macros/relations/table/create.sql#L2

```
{% macro sqlserver__create_table_as(temporary, relation, sql) -%}
    {%- set query_label = apply_label_for_table() -%}
    ...

```

Benjamin-Knight:
Supports a dict form (allowlisted hints; MAX_GRANT_PERCENT / MIN_GRANT_PERCENT use `=` syntax) and a query_options_raw escape hatch. Usable on table, incremental (delete+insert, merge, microbatch), snapshot, and unit_test materializations.
Usage on a view raises a compiler error if set. 
Apply_label() is preserved as a backward-compat alias emitting only the LABEL hint.
Fixes #613

### DIFF
--- a/dbt/include/sqlserver/macros/adapters/metadata.sql
+++ b/dbt/include/sqlserver/macros/adapters/metadata.sql
@@ -26,7 +26,6 @@
             'MAXRECURSION',
             'NO_PERFORMANCE_SPOOL',
             'OPTIMIZE FOR UNKNOWN',
-            'PARAMETERIZATION',
             'QUERYTRACEON',
             'RECOMPILE',
             'ROBUST PLAN',
@@ -46,11 +45,18 @@
                     {{ exceptions.raise_compiler_error("Query option '" ~ key ~ "' value must be a number, got: '" ~ value ~ "'") }}
                 {%- endif -%}
                 {%- set separator = ' = ' if key | upper in equals_syntax_options else ' ' -%}
-                {%- do options_list.append(key | upper ~ separator ~ value | int) -%}
+                {#- Render the value verbatim: ints become "1", floats become "12.5".
+                    MAX_GRANT_PERCENT / MIN_GRANT_PERCENT accept decimals 0.0–100.0; integer-only
+                    options will surface a clear SQL Server parse error on invalid decimals. -#}
+                {%- do options_list.append(key | upper ~ separator ~ value) -%}
             {%- endif -%}
         {%- endfor -%}
 
-        {#- query_options_raw bypasses the allowlist; users opt in to writing valid SQL Server syntax themselves. -#}
+        {#- query_options_raw bypasses the allowlist; users opt in to writing valid SQL Server syntax themselves.
+            Shape-check only: a plain string would be iterated character-by-character into garbage. -#}
+        {%- if query_options_raw is string or query_options_raw is mapping -%}
+            {{ exceptions.raise_compiler_error("query_options_raw must be a list of strings, got: '" ~ query_options_raw ~ "'") }}
+        {%- endif -%}
         {%- for raw in query_options_raw -%}
             {%- do options_list.append(raw) -%}
         {%- endfor -%}
@@ -59,17 +65,16 @@
     OPTION ({{ options_list | join(', ') }});
 {% endmacro %}
 
-{#- Backward-compat alias for the pre-1.10 macro. Emits only the LABEL hint
-    and ignores query_options / query_options_raw. New adapter code should
-    call get_query_options() directly.
+{#- DEPRECATED: backward-compat alias for the pre-1.10 macro.
 
-    Note: this preserves non-breaking *consumption* of apply_label (user
-    macros calling `{{ apply_label() }}` still resolve), but does NOT
-    preserve non-breaking *override*: adapter macros no longer call
-    apply_label internally, so a project that overrides apply_label in its
-    own macros directory will find that override has no effect on adapter
-    behaviour. To customise the OPTION clause emitted by adapter macros,
-    override get_query_options instead. -#}
+    Calls to `{{ apply_label() }}` from user macros still resolve and emit
+    a LABEL-only OPTION clause — but apply_label() is no longer the
+    extension point. Adapter macros now call get_query_options() instead,
+    so overriding apply_label() in a project's macros directory will have
+    no effect on adapter-emitted SQL.
+
+    To customise the OPTION clause emitted by adapter macros (table,
+    incremental, snapshot, unit_test), override get_query_options instead. -#}
 {% macro apply_label() %}
     {{ log (config.get('query_tag','dbt-sqlserver'))}}
     {%- set query_label = config.get('query_tag','dbt-sqlserver') -%}
--- a/dbt/include/sqlserver/macros/adapters/metadata.sql
+++ b/dbt/include/sqlserver/macros/adapters/metadata.sql
@@ -1,7 +1,90 @@
+{% macro get_query_options(parse_options=False) %}
+    {{ log (config.get('query_tag','dbt-sqlserver'))}}
+    {%- set query_label = config.get('query_tag','dbt-sqlserver') -%}
+    {%- set query_options = config.get('query_options', {}) -%}
+    {%- set query_options_raw = config.get('query_options_raw', []) -%}
+
+    {%- set options_list = ["LABEL = '" ~ query_label ~ "'"] -%}
+
+    {%- if parse_options -%}
+        {%- set valid_options = [
+            'HASH GROUP', 'ORDER GROUP',
+            'CONCAT UNION', 'HASH UNION', 'MERGE UNION',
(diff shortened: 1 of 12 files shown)


dbt/include/sqlserver/macros/adapters/metadata.sql
```diff
@@ -1,9 +1,92 @@
...
+    and ignores query_options / query_options_raw. New adapter code should
+    call get_query_options() directly.
+
+    Note: this preserves non-breaking *consumption* of apply_label (user
+    macros calling `{{ apply_label() }}` still resolve), but does NOT
+    preserve non-breaking *override*: adapter macros no longer call
+    apply_label internally, so a project that overrides apply_label in its
+    own macros directory will find that override has no effect on adapter
+    behaviour. To customise the OPTION clause emitted by adapter macros,
+    override get_query_options instead. -#}
```
axellpadilla: Please add this deprecation on the changelog so we don't forget to publish it on release or check if we can avoid it as a standard execution path, this can reasonably fail with a targeted compiler error and instruct them to override get_query_options() instead. 
Should we make something like this?:
- Make get_query_options() compose with apply_label() so existing apply_label() overrides still affect generated SQL.
- Add a test where a project overrides apply_label() and verifies table/incremental SQL still includes the custom label.
- Document get_query_options() as the new extension point for full query-hint customization.
- Keep apply_label() supported for at least one release cycle?

dbt/include/sqlserver/macros/adapters/metadata.sql
```diff
@@ -1,9 +1,92 @@
...
+            {%- if value is none -%}
+                {%- do options_list.append(key | upper) -%}
+            {%- else -%}
+                {%- if value is not number -%}
+                    {{ exceptions.raise_compiler_error("Query option '" ~ key ~ "' value must be a number, got: '" ~ value ~ "'") }}
+                {%- endif -%}
+                {%- set separator = ' = ' if key | upper in equals_syntax_options else ' ' -%}
+                {%- do options_list.append(key | upper ~ separator ~ value | int) -%}
+            {%- endif -%}
+        {%- endfor -%}
```
axellpadilla, line 53: One issue with the current allowlist/rendering logic: all numeric values are emitted with `value | int`, which silently changes valid decimal values for `MAX_GRANT_PERCENT` / `MIN_GRANT_PERCENT`.
https://learn.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-query?view=sql-server-ver17

For example:

query_options:
  MAX_GRANT_PERCENT: 12.5

would render as:

OPTION (LABEL = '...', MAX_GRANT_PERCENT = 12)

SQL Server allows these grant-percent hints to use decimal numeric values between `0.0` and `100.0`, so we should either preserve decimals for those options or remove them from the structured allowlist and require `query_options_raw` for them.

Related: `PARAMETERIZATION` is currently allowlisted, but it cannot render valid syntax with the current value handling. SQL Server expects:

PARAMETERIZATION { SIMPLE | FORCED }

So `PARAMETERIZATION: FORCED` currently fails because strings are rejected, while `PARAMETERIZATION: null` would render invalid SQL. I think we should either support a string enum of `SIMPLE` / `FORCED`, or remove `PARAMETERIZATION` from the allowlist and require `query_options_raw` for it.

dbt/include/sqlserver/macros/adapters/metadata.sql
```diff
@@ -1,9 +1,92 @@
+{% macro get_query_options(parse_options=False) %}
+    {{ log (config.get('query_tag','dbt-sqlserver'))}}
+    {%- set query_label = config.get('query_tag','dbt-sqlserver') -%}
+    {%- set query_options = config.get('query_options', {}) -%}
+    {%- set query_options_raw = config.get('query_options_raw', []) -%}
```
axellpadilla, line 5: `query_options_raw` should probably validate its input shape. Since it is intended to be a list of raw query-option strings, passing a plain string can produce confusing output because Jinja may iterate over the string character by character.

For example:
query_options_raw: "USE HINT('DISABLE_OPTIMIZER_ROWGOAL')"
could render as separate characters instead of one option item.

Suggested fix: add a compiler error when `query_options_raw` is a string, and ideally validate that every item in the list is also a string. This would make misconfiguration fail clearly instead of producing invalid SQL.

tests/functional/adapter/mssql/test_query_options.py
axellpadilla, line 1: - decimal values for `MAX_GRANT_PERCENT` / `MIN_GRANT_PERCENT` should not be silently truncated by `value | int` (if we choose to support it)
- `PARAMETERIZATION` should either render valid `SIMPLE` / `FORCED` syntax or be rejected from the structured allowlist
- `query_options_raw` should reject a plain string instead of iterating over it character by character

Most of these don’t need to hit SQL Server. A unit-style test that compiles/renders the macro output and asserts the generated SQL or compiler error should be enough. Full functional coverage could remain limited to one happy-path model/materialization test.

axellpadilla (changes requested):

dbt/include/sqlserver/macros/adapters/metadata.sql
Benjamin-Knight: I decided against trying to parse the raw option, but didn't really have anything to state why, its more of an escape hatch for those who want to do more complicated thing that the basic options can't cover, as you can see with the other issues trying to safely handle every single SQL option is painful so I went for a smaller subset, plus I don't want constant pull requests to add in support for an option.
As such this raw option is very much on the user to get right, if they provide invalid options then the SQL they get is invalid.

axellpadilla: Makes sense, we still should just validate is indeed a list before processing as a list, or, support raw text?

dbt/include/sqlserver/macros/adapters/metadata.sql
Benjamin-Knight: I'm not sure how we could keep apply label support. If anyone was using apply label overrides to apply their own options this would supersede it. The best we could do is make apply labels call the new macro internally and just print the label as it did before, that means callers of apply_label would work as before, but that is the same effect as just leaving apply_label as is, just without the duplication of code.

The risk surface is users who are overriding apply_label. If they are over riding it to do something other than applying a label because it was a useful injection point then they will have a migration issue, but I cannot see a way around that other than notifying them of the new path. I wouldn't call it a breaking change, as its a not a proper usage of the macro.

axellpadilla (approved):

tests/functional/adapter/mssql/test_query_options.py
```diff
@@ -0,0 +1,844 @@
...
+        results = run_dbt(["run"])
+        assert len(results) == 1
+        assert results[0].status == "success"
+
+        sql = _find_compiled_run_sql(project, "inc_model.sql")
+        assert "MAXDOP 1" in sql
+
+
+# ---------------------------------------------------------------------------
+# Incremental merge / microbatch — unsupported, must raise
```
axellpadilla:
```suggestion
# Incremental merge / microbatch
```

axellpadilla: Looks like it is?

PR #681, opened 2026-05-21
Feat/mssql python Add optional mssql-python backend and update docs / CI

axellpadilla:
This PR introduces the new optional mssql-python connection backend while keeping the legacy pyodbc ODBC path available as an explicit extra.

Key changes:

- Add lazy-loaded backend support in [sqlserver_connections.py](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
- Make both pyodbc and mssql-python optional extras in [pyproject.toml](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
- Update [sqlserver_credentials.py](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) for backend selection
- Add/adjust unit tests for lazy import and backend behavior
- Update integration test matrix in [integration-tests-sqlserver.yml](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to preserve coverage without increasing total run count
- Add devcontainer support for both backend extras in [setup_env.sh](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
- Clarify installation and usage in [README.md](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
- Add developer setup notes and dependency guidance to [CONTRIBUTING.md](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
- Small support changes in [test.env.sample](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [conftest.py](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), and [uv.lock](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

Why this change:
- enables the modern mssql-python backend behind currently gated use_mssql_python: true
- Avoids requiring system dependencies unless a backend is explicitly selected
- keeps existing pyodbc behavior intact for legacy profiles
- improves documentation and local development setup

Testing:
- updated unit tests for both backend import behavior
- CI workflow reduces repeated tests and covers partitioning backends across Python versions
- devcontainer setup installs both pyodbc and mssql-python extras for local validation


Benjamin-Knight:
Ok, its a big change so will take a while to review, I can see a few issues I think so I'll find some time at start providing feedback.

My main concern other than the actual code change is the install method for the adapter now changes as we have to specify the driver we want to use, that is a big change. This doesn't just affect us, DBT themselves will have the wrong install instructions.

https://docs.getdbt.com/docs/local/connect-data-platform/mssql-setup?version=1.12#prerequisites

dbt/adapters/sqlserver/sqlserver_connections.py
```diff
@@ -72,6 +77,68 @@ class AccessToken:  # type: ignore[no-redef]
     "decimal.Decimal": "decimal",
 }
 
+MSSQL_PYTHON_UNSUPPORTED_AUTHENTICATIONS = {
```
Benjamin-Knight: You do not have MSI in this list, there is no MSI code in the new adapter connection handling, its only in get_pyodbc_attrs_before_credentials.

dbt/adapters/sqlserver/sqlserver_connections.py
```diff
@@ -353,6 +454,132 @@ def byte_array_to_datetime(value: bytes) -> dt.datetime:
...
+) -> Tuple[Type[Exception], ...]:
+    if _use_mssql_python_backend(credentials):
+        mssql_python = _get_mssql_python()
+
+        retryable_exceptions = [
+            getattr(mssql_python, "InternalError", Exception),
+            getattr(mssql_python, "OperationalError", Exception),
+        ]
+
+        if _requires_pyodbc_backend(credentials):
```
Benjamin-Knight: This is inside a mssql python branch but states its pyodbc, I think the _requires_pyodbc_backend just needs renaming, I think this might have been an attempt to standardize logic between two branches but the same logic appears without the function on line 573 as well? Probably needs calling there as well?

dbt/adapters/sqlserver/sqlserver_connections.py
```diff
@@ -353,6 +454,132 @@ def byte_array_to_datetime(value: bytes) -> dt.datetime:
...
+    return f"{credentials.host},{credentials.port}"
+
+
+def _build_common_connection_string_parts(
+    credentials: SQLServerCredentials,
+) -> list[str]:
+    con_str = [f"SERVER={_build_server_arg(credentials)}"]
+    con_str.append(f"Database={credentials.database}")
+
+    assert credentials.authentication is not None
```
Benjamin-Knight: These assert statements, so here and L494-495 are not safe, if run in optimization mode they are dropped and these values could end up resolving to false or something unexpected, just raise a DBT exception of some kind.

dbt/adapters/sqlserver/sqlserver_credentials.py
```diff
@@ -6,10 +6,10 @@
 
 @dataclass
 class SQLServerCredentials(Credentials):
-    driver: str
-    host: str
-    database: str
-    schema: str
+    driver: Optional[str] = None
+    host: str = ""
```
Benjamin-Knight: These string defaults do not look safe, previously not provided a value would cause a proper error before connection, now it will just try and use invalid blank strings.

pyproject.toml
```diff
@@ -40,11 +40,16 @@ azure = [
 pyodbc = [
     "pyodbc>=5.2.0",
 ]
+mssql = [
+    "mssql-python>=1.1.0",
```
Benjamin-Knight: Why such an old version of the driver? We need at least 1.4.0 for the fixes to the datetime handling.

dbt/adapters/sqlserver/sqlserver_connections.py
```diff
@@ -392,101 +633,45 @@ def open(cls, connection: Connection) -> Connection:
...
 
-        con_str_display = ";".join(con_str)
+            if _use_mssql_python_backend(credentials):
+                assert mssql_python is not None
 
-        retryable_exceptions = [  # https://github.com/mkleehammer/pyodbc/wiki/Exceptions
-            pyodbc.InternalError,  # not used according to docs, but defined in PEP-249
-            pyodbc.OperationalError,
-        ]
+                mssql_python.pooling(enabled=False)
```
Benjamin-Knight: pyodbc uses pooling, why are we defaulting to no pooling on mssql_python?

.github/workflows/integration-tests-sqlserver.yml
```diff
@@ -35,17 +35,39 @@ on:  # yamllint disable-line rule:truthy
...
     strategy:
       matrix:
         python_version: ["3.10", "3.11", "3.12", "3.13"]
         msodbc_version: ["17", "18"]
         sqlserver_version: ["2017", "2019", "2022"]
         collation: ["SQL_Latin1_General_CP1_CS_AS", "SQL_Latin1_General_CP1_CI_AS"]
+        backend:
+          - pyodbc
+          - mssql-python
+        exclude:
```
Benjamin-Knight, line 50: pyodbc is still going to be the default for most users, at least to start, I do not think we should drop python version testing from pyodbc, I'm okay with the recent python versions only for the mssql-python, if you are changing driver you are likely on up to date python.

dbt/adapters/sqlserver/sqlserver_connections.py
```diff
@@ -353,6 +454,132 @@ def byte_array_to_datetime(value: bytes) -> dt.datetime:
...
+    if credentials.trace_flag:
+        con_str.append("SQL_ATTR_TRACE=SQL_OPT_TRACE_ON")
+    else:
+        con_str.append("SQL_ATTR_TRACE=SQL_OPT_TRACE_OFF")
+
+    plugin_version = __version__.version
+    application_name = f"dbt-{credentials.type}/{plugin_version}"
+    con_str.append(f"APP={application_name}")
+
+    try:
```
Benjamin-Knight: Don't think we need a try except around list appends?

dbt/adapters/sqlserver/sqlserver_connections.py
```diff
@@ -84,6 +151,40 @@ def _require_azure_identity(authentication: str) -> None:
...
+    authentication = str(credentials.authentication or "sql").lower().strip()
+    return authentication in AZURE_AUTH_FUNCTIONS or authentication == "activedirectoryaccesstoken"
+
+
+def _use_mssql_python_backend(credentials: SQLServerCredentials) -> bool:
+    return bool(getattr(credentials, "use_mssql_python", False))
+
+
+def _validate_pyodbc_requirements(credentials: SQLServerCredentials) -> None:
+    if not credentials.driver:
```
Benjamin-Knight: This would allow someone providing a driver value of "", we probably want something stronger like. 
if credentials.driver is None or not credentials.driver.strip():

tests/unit/adapters/mssql/test_sqlserver_connection_manager.py
```diff
@@ -65,3 +71,238 @@ def test_get_pyodbc_attrs_before_cli_auth_requires_azure_identity(
...
+
+    con_str = captured["connection_string"]
+    assert "DRIVER=" not in con_str
+    assert "SERVER=fake.sql.sqlserver.net,1433" in con_str
+    assert "Database=dbt" in con_str
+    assert "UID={dbt_user}" in con_str
+    assert "PWD={super-secret}" in con_str
+    assert "encrypt=Yes" in con_str
+    assert "TrustServerCertificate=Yes" in con_str
+    assert "APP=dbt-sqlserver/" not in con_str
```
Benjamin-Knight, line 1261: Why are we testing that the APP value is no longer set? We set it in pyodbc and its useful when debugging server side, does the new driver not support it? Its a common connection string value? I can see we do not try and set it in the new driver connection code.

Benjamin-Knight (changes requested):
Probably need some new tests as well but will look at those after these comments.

axellpadilla:
@Benjamin-Knight I'm moving this to v1.11, I'm more interested in the other patches for v1.10 and its true we will need more time holding the rc1 pending until dbt docs are updated

.github/workflows/integration-tests-sqlserver.yml
axellpadilla: I just dont want to increase the testing times by a lot, how could we divide this? maybe by collation?

pyproject.toml
axellpadilla: This trial is old, will set to 1.7.1 to be able to support msi

tests/unit/adapters/mssql/test_sqlserver_connection_manager.py
axellpadilla: this is not compatible with mssql becasue "Reserved keyword 'app' is controlled by the driver and cannot be specified by the user"

dbt/adapters/sqlserver/sqlserver_connections.py
axellpadilla: I had errors on the old version

dbt/adapters/sqlserver/sqlserver_connections.py
axellpadilla: implemented using new version dont have a way to test it tough

.github/workflows/integration-tests-sqlserver.yml
Benjamin-Knight: We could consider trimming SQL 2017, its in extended support for another year, not sure how often people are using it. 
We could consider testing different collations of SQL server under recent versions of python only. Really the collation tests although we may need to do a pass on the interaction points between python and SQL to see how risky that would be, I know date and times are a friction point point but I'm not sure collations are which means that the collations are only really needed once to make sure our SQL code is correct.

axellpadilla: just sent a new one that should cover all compatibility cases without being exagerated, let me know what you think

Benjamin-Knight: Yes looks good, covers all the cases I think we need.

axellpadilla:
@Benjamin-Knight Kept pyodbc by default and already running pyodbc part on production so changed milestone to v1.10.1, this way we don't need to update upstream docs still, and can announce deprecation if we really start to mostly use the new driver.
All lazy imports behavior intact, this mostly makes a mandatory install of pyodbc by default even if we don't use it.
This should be squash merged after v1.10.0

Benjamin-Knight:
> @Benjamin-Knight Kept pyodbc by default and already running pyodbc part on production so changed milestone to v1.10.1, this way we don't need to update upstream docs still, and can announce deprecation if we really start to mostly use the new driver. All lazy imports behavior intact, this mostly makes a mandatory install of pyodbc by default even if we don't use it. This should be squash merged after v1.10.0

Ok, Its a big change so we will want to have a much longer period for a release candidate then we did with the 1.9.1 release to give people time to feedback but I think its good to go now that the default behaviour is still pyodbc.

Have you flipped the adapter into mssql driver mode for a large corpus of models?

Benjamin-Knight (approved):

axellpadilla:
> > @Benjamin-Knight Kept pyodbc by default and already running pyodbc part on production so changed milestone to v1.10.1, this way we don't need to update upstream docs still, and can announce deprecation if we really start to mostly use the new driver. All lazy imports behavior intact, this mostly makes a mandatory install of pyodbc by default even if we don't use it. This should be squash merged after v1.10.0
> 
> Ok, Its a big change so we will want to have a much longer period for a release candidate then we did with the 1.9.1 release to give people time to feedback but I think its good to go now that the default behaviour is still pyodbc.
> 
> Have you flipped the adapter into mssql driver mode for a large corpus of models?

Going to on a couple days of pyodbc, current side effect is a speed gain on sqlfluff formatting because the lazy stuff

PR #685, opened 2026-05-25
✨ feat: add default schema concatenation flag support and update documentation

### LINKED ISSUE #683: Migrate `dbt_sqlserver_use_default_schema_concat` to adapter behavior flag
### Context

`dbt_sqlserver_use_default_schema_concat` is currently documented in the README under flags, but it is configured through `vars`:

```yaml
vars:
  dbt_sqlserver_use_default_schema_concat: true
````

This controls whether SQL Server keeps legacy schema behavior or delegates to dbt-core’s default schema concatenation.

### Problem

This setting behaves like an adapter behavior flag, but it is not currently exposed through dbt’s standard `flags:` configuration.

This is confusing because users see it documented as a flag, but must configure it as a project variable.

### Proposal

Add support for the standard adapter behavior flag approach:

```yaml
flags:
  <adapter_flag_name>: true
```

Suggested flag name:

```yaml
flags:
  dbt_sqlserver_use_default_schema_concat: true
```

Keep the current `vars:` configuration temporarily for backwards compatibility until flag deprecation:

```yaml
vars:
  dbt_sqlserver_use_default_schema_concat: true
```

### Migration behavior

* Prefer the new adapter behavior flag when present.
* Continue supporting the existing `vars:` setting during the transition.
* Emit a warning when the legacy `vars:` setting is used.
* Document the migration path in the README.
* Plan deprecation/removal of the `vars:` fallback for `v1.11.0`.

### Why

The setting is a behavior switch between:

* legacy SQL Server behavior: `custom_schema_name`
* dbt-core/upstream behavior: `<target.schema>_<custom_schema_name>`

This matches dbt’s adapter behavior-flag model more closely than a project variable.

### Related context

This became more visible after the Fabric dependency was removed in `dbt-sqlserver==1.9.1`.

With:

```txt
dbt-sqlserver==1.9.0
dbt-fabric>=1.9.3
```

users effectively got dbt-core schema behavior through Fabric after its schema override was removed:

[https://github.com/microsoft/dbt-fabric/issues/266](https://github.com/microsoft/dbt-fabric/issues/266)

After `dbt-sqlserver==1.9.1`, users need to explicitly enable the compatibility setting to keep that behavior.

### Release plan

- `v1.9.2`: add adapter behavior flag support, keep `vars:` fallback, emit warnings
- `v1.10.x`: continue compatibility period
- `v1.11.0`: remove or deprecate the `vars:` and "flag" and the schema override by the adapter

axellpadilla:
Closes #683
Should be backported to v1.9.2

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,11 +68,13 @@ See [the changelog](CHANGELOG.md)
   **Example usage in `dbt_project.yml`:**
 
   ```yaml
-  vars:
+  flags:
     dbt_sqlserver_use_default_schema_concat: true  # Enable standard schema concatenation
   ```
 
-  > **Note:** If you want to permanently customise schema generation and avoid any future deprecation of this flag, override the `sqlserver__generate_schema_name` macro directly in your project.
+  This adapter also supports the same setting via `vars:` for backwards compatibility, so either method works in the current release.
+
+  > **Note:** If you want to permanently customise schema generation and avoid any future changes, override the `sqlserver__generate_schema_name` macro directly in your project instead.
 
 
 
--- a/dbt/include/sqlserver/macros/adapters/schema.sql
+++ b/dbt/include/sqlserver/macros/adapters/schema.sql
@@ -48,16 +48,19 @@
       - `target.schema`              when no custom schema is set
       - `custom_schema_name` (trim)  when a custom schema is set
 
-    When the `dbt_sqlserver_use_default_schema_concat` variable is set to true,
+    When the `dbt_sqlserver_use_default_schema_concat` flag is enabled,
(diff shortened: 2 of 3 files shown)


axellpadilla:
@Benjamin-Knight is there any other fix do you want to include in v1.9.2 to backport them?

Benjamin-Knight:
> @Benjamin-Knight is there any other fix do you want to include in v1.9.2 to backport them?

No, just bug fixes for the the 1.9 branch and this one qualifies as one, the interaction between fabric changing this back and forth and our setup for controlling schema behaviour is not intuitive and this helps.

PR #692, opened 2026-06-02
Update version to 1.10.0rc1 on release/v1.10

axellpadilla:


joshmarkovic:
Thank you to both @axellpadilla and @Benjamin-Knight!!

PR #693, opened 2026-06-02
Update changelog for v1.10.0

axellpadilla:


Benjamin-Knight:
We are missing details around #603, #610, #686
We should probably call out the dropping of 3.9 support and the PEP work for anyone who does download the project itself.

PR #699, opened 2026-06-04
Empty unit test fixtures (rows: []) emit invalid LIMIT 0 syntax

### LINKED ISSUE #698: Bug: rows: [] in unit test given block emits LIMIT 0
Adapter: dbt-sqlserver 1.10.0-rc1
dbt Core: 1.10.20
SQL Server: Microsoft ODBC Driver 18

### Description

When a unit test given block specifies rows: [] for an input, dbt core generates an
empty-fixture CTE of the form:

```
__dbt__cte__some_ref as (
    -- Fixture for some_ref
    select try_cast(null as varchar(255)) as "col_a"
    limit 0
)
```

The query fails immediately with:

('42000', "[42000] [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]
Incorrect syntax near 'limit'. (102) (SQLMoreResults)")

### Steps to reproduce

1. Define a dbt unit test where at least one given input has rows: [].
2. Run dbt test --resource-type unit_test.

Minimal YAML excerpt:

```
unit_tests:
  - name: my_test
    model: my_model
    given:
      - input: ref('some_ref')
        rows: []          # <-- triggers the bug
      - input: ref('other_ref')
        rows:
          - col_a: 'foo'
    expect:
      rows:
        - col_a: 'foo'
```

### Expected behaviour

An empty rows: [] block should produce a zero-row CTE using T-SQL-compatible
syntax

Benjamin-Knight:
Fixes #698

dbt-core hardcodes `limit 0` for empty fixture rows, which is not valid T-SQL. These macros are not dispatched, so shadow them in the adapter package and emit `select top 0` instead. 

Also fix sqlserver__get_columns_in_query for queries starting with a CTE (hit by unit tests with an empty `expect` block): such queries cannot be wrapped in a subquery, so describe their result set via sp_describe_first_result_set instead of executing them. Non-CTE queries keep the existing TOP 0 wrapping. The CTE detection is factored into a shared sqlserver__select_starts_with_cte helper.


dbt/adapters/sqlserver/__version__.py
axellpadilla, line 1: Please base your changes from master if aiming v1.10.1 to avoid the temporal rc1 change I made only into the release branch

dbt/include/sqlserver/macros/materializations/models/unit_test/get_fixture_sql.sql
axellpadilla, line 1: I don't see any other solution than shadowing, that said, inside the shadowed macro we could introduce a call to macro sqlserver__get_fixture_sql(...) instead. This would keep the SQL Server-specific logic isolated and make the implementation closer to what we would want upstream if they use adapter calls later.

This is not  that important but ideally with a link to an upstream dbt-core/dbt-adapters issue requesting dispatch support. If upstream adds that dispatch point later, our future patch should mostly be removing the shadow and keeping only the SQL Server-specific sqlserver__get_fixture_sql implementation.

axellpadilla (changes requested):

dbt/adapters/sqlserver/__version__.py
Benjamin-Knight: Think I had already done this?

axellpadilla:
Referencing issue https://github.com/dbt-labs/dbt-adapters/issues/2021

axellpadilla (approved):

PR #710, opened 2026-06-15
Implements `dbt_sqlserver_use_dbt_transactions` behavior flag

### LINKED ISSUE #708: Support opt-in dbt-managed SQL Server transactions
## Problem

`dbt-sqlserver` currently opens SQL Server connections with driver `autocommit=True`, and adapter-level `add_begin_query()` / `add_commit_query()` are no-ops.

This has been the adapter behavior for a long time. There is historical commit context showing autocommit was enabled to mitigate “weird behavior,” and later `BEGIN TRANSACTION` / `COMMIT TRANSACTION` were removed for autocommit. However, I could not find a more detailed design note explaining whether the current no-op transaction model is still required.

This means dbt’s transaction flow is mostly logical for this adapter. “Inside transaction” hooks and `adapter.commit()` do not necessarily map to a real SQL Server transaction unless a macro explicitly emits `BEGIN TRANSACTION` / `COMMIT TRANSACTION`.

## Proposal

Add an opt-in behavior flag option for dbt-managed SQL Server transactions to go back to dbt standard en enable correct transaction handling, for example:

```yaml
dbt_sqlserver_use_dbt_transactions: false
````

Default should remain `false`, preserving current behavior.

When disabled:

* keep `add_begin_query()` / `add_commit_query()` as SQL no-ops
* preserve existing adapter behavior

When enabled:

* make `add_begin_query()` emit `BEGIN TRANSACTION`
* make `add_commit_query()` emit `COMMIT TRANSACTION`
* add/verify rollback behavior, preferably guarded with `IF @@TRANCOUNT > 0 ROLLBACK TRANSACTION`

## Why

Autocommit itself is not the only problem. The bigger issue is that dbt’s begin/commit hooks are currently fictitious at the SQL Server level.

An opt-in explicit transaction mode would let us evaluate real dbt transaction semantics without changing behavior for existing projects.

## Suggested scope

* Add a new option, e.g. `dbt_sqlserver_use_dbt_transactions`, defaulting to `false`.
* When `dbt_sqlserver_use_dbt_transactions: true`, implement:

  * `add_begin_query()` -> `BEGIN TRANSACTION`
  * `add_commit_query()` -> `COMMIT TRANSACTION`
  * rollback behavior -> `IF @@TRANCOUNT > 0 ROLLBACK TRANSACTION`

* Add tests for:

  * default mode unchanged
  * no duplicate `BEGIN TRANSACTION` in normal materialization flow
  * regular pre/post hooks inside a transaction
  * table materialization success/failure
  * incremental materialization success/failure
  * rollback/cleanup after failed model builds
  * subsequent model can run after a failed model
  * explicit transaction SQL in user hooks/macros is either supported or clearly documented as unsupported in this mode

## Acceptance criteria

* Existing behavior remains unchanged by default.
* New opt-in mode emits real SQL Server transaction statements through dbt’s adapter transaction hooks.
* Functional tests cover success, failure, rollback/cleanup, and hook behavior.

axellpadilla:
Adds an opt-in flag that makes dbt's transaction hooks real at the SQL Server level, replacing the current no-op behavior, couldn't find the exact reason this was removed and this stopsome issues we currently have with left over temp tables, dirty reads of empty tables, non-standard behaviors.

### What it does

When `dbt_sqlserver_use_dbt_transactions: true` is set in `dbt_project.yml`:

- **`add_begin_query()`** emits `BEGIN TRANSACTION` instead of being a no-op
- **`add_commit_query()`** emits `IF @@TRANCOUNT > 0 COMMIT TRANSACTION` instead of being a no-op
- **`_rollback_handle()`** emits `IF @@TRANCOUNT > 0 ROLLBACK TRANSACTION` with error handling matching the base class pattern (fires `RollbackFailed` event on failure)
- **After-commit hooks** — the run_hooks macro uses the flag to decide whether hooks run after a real commit or after the previous autocommit-based cleanup

When `false` (default), behavior is unchanged — `begin`/`commit` remain no-ops and each statement is auto-committed by the ODBC driver.

### Why opt-in

Enabling real transactions changes the atomicity guarantees and maps to default dbt standard behavior. Without the flag, each SQL statement commits individually — a failure mid-materialization doesn't roll back earlier DDL/DML. With the flag, the entire model run is wrapped in a single T-SQL transaction, and failures roll back everything.

This includes the DML refresh materialization (`table_dml_refresh.sql`), which has its own explicit `BEGIN TRANSACTION` for the atomic DELETE+INSERT swap. With the flag on, this nests inside dbt's outer transaction. SQL Server handles this correctly — only the outermost COMMIT actually commits, and a rollback always rolls back all nesting levels — so scratch tables from a partially-executed materialization are cleaned up on failure.

### Other changes

- Updated the stale comment in `table_dml_refresh.sql` to reflect the flag
- Fixed a line length issue in test files
- Added functional tests for table/view/incremental materializations, rollback behavior (with and without the flag), after-commit post-hooks, and recovery after a failed model including xfail temporal variants that will never pass without the flag.
- Added unit tests for `add_begin_query`/`add_commit_query` flag gating

Closes #708

### DIFF
--- a/dbt/include/sqlserver/macros/materializations/hooks.sql
+++ b/dbt/include/sqlserver/macros/materializations/hooks.sql
@@ -2,7 +2,11 @@
   {% for hook in hooks | selectattr('transaction', 'equalto', inside_transaction)  %}
     {% if not inside_transaction and loop.first %}
       {% call statement(auto_begin=inside_transaction) %}
-        if @@trancount > 0 commit;
+        {% if not adapter.behavior.dbt_sqlserver_use_dbt_transactions %}
+          if @@trancount > 0 commit; -- post hooks after fictitious transaction work as expected
+        {% else %}
+          commit; -- align transaction=False hook behavior with dbt-core transaction semantics.
+        {% endif %}
       {% endcall %}
     {% endif %}
     {% set rendered = render(hook.get('sql')) | trim %}
--- a/README.md
+++ b/README.md
@@ -129,6 +129,24 @@ The same setting is also honoured via `vars:` for backwards compatibility; the b
 
 *(default: `pyodbc`)* Set to `mssql-python` in a profile target to use the `mssql-python` backend instead of `pyodbc`. The adapter fails if the required backend package (Python dependency), such as `pyodbc` or `mssql-python`, is not installed.
 
+### `dbt_sqlserver_use_dbt_transactions`
+
+_(default: `false`)_ When enabled, makes dbt's transaction hooks real at the SQL Server level by emitting `BEGIN TRANSACTION` / `COMMIT TRANSACTION` through the adapter's `add_begin_query` and `add_commit_query` methods. 
+
+The default is `false`, preserving existing behavior where `begin`/`commit` hooks are logical no-ops and the ODBC driver auto-commits each statement. When `dbt_sqlserver_use_dbt_transactions: true`, the adapter emits real T-SQL transaction statements, and rollback uses `IF @@TRANCOUNT > 0 ROLLBACK TRANSACTION`.
+
+The driver connection remains in autocommit mode (`autocommit=true`) in both modes.
+
+This mode is opt-in and should be tested carefully with project-specific materializations and hooks.
+
+```yaml
+# dbt_project.yml
+flags:
+    dbt_sqlserver_use_dbt_transactions: true # <-- opt-in; default is false
+```
+
+**Compatibility notes:** Enabling `dbt_sqlserver_use_dbt_transactions: true` may expose transaction-state assumptions hidden by autocommit-only mode. Explicit transaction macros may interact with dbt-managed transactions, and cleanup after failed DDL/DML may differ. Review pre/post hooks for in-transaction vs out-of-transaction semantics.
+
 ## Contributing
 
 [![Unit tests](https://github.com/dbt-msft/dbt-sqlserver/actions/workflows/unit-tests.yml/badge.svg)](https://github.com/dbt-msft/dbt-sqlserver/actions/workflows/unit-tests.yml)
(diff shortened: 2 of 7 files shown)


dbt/include/sqlserver/macros/materializations/hooks.sql
Benjamin-Knight, line 1: hooks.sql reads the flag from a different source (flags) than the connection manager (adapter.behavior), and the two agree only while the default is False

Benjamin-Knight (commented):
The DML work I did has an explicit transaction in it, this probably should only be emitted when this flag is false because when DBT is handling the transaction all it does it create a nested transaction that doesn't add anything to the process.

I can't see the XACT_ABORT work that @joshmarkovic identified as being an issue. If we want DBT to handle transaction should be have this flag emit the XACT_ABORT setting as well? 

I think we need a test to cover snapshots as they call adapter.commit in their code path.

Benjamin-Knight:
I remember someone stating that there are parts of DBT which we may want to not run inside the transaction which is now DBT managed, mostly post hooks, are we planning on shipping this without that resolved? Right now this work would break online and resumable post hook indices that are declared.

axellpadilla:
@Benjamin-Knight noted the other comments to check. About the parts of dbt, because post hooks that should be resumable, should be stated as not in transaction?, this feature was fixed in #709, and probably nobody is using it, we were doing no transaction practices but because it was not posible, this feature will probably remain optional for a long time (probably only defaulted in v2.1.0?) and recommendations of better future practices should be documented.

axellpadilla:
@Benjamin-Knight  I agree XACT_ABORT is related, but I would treat it as a follow-up. This PR makes dbt transaction boundaries real; XACT_ABORT changes how SQL Server reacts to runtime errors inside those boundaries. If we add it, it should be explicit, tested, and probably scoped/restored carefully, lest better isolate that change. Fixed other comments and did a rework to formalize final state and commets. Thanks.

PR #714, opened 2026-06-16
#535 Postgres style indexes

### LINKED ISSUE #535: Implement Postgres style index creation
Add index creation structure like dbt-postgres through an index configuration statement.

https://docs.getdbt.com/reference/resource-configs/postgres-configs#indexes

Benjamin-Knight:
Based off @axellpadilla original work, been using myself for a couple of weeks so appears stable, I was about to base this on the #710 work due to the transactions and online/resumable indices but instead have made it independent of that and just moved them out of the transaction completely.

### Postgres-style indexes model config for SQL Server

  Closes #535.

  Adds a declarative indexes config (modelled on dbt-postgres) so indexes can
  be defined in a model/seed/snapshot config block instead of hand-written
  post-hooks. dbt owns the full lifecycle: it creates them on build,
  reconciles them against the config on subsequent runs, and never touches
  indexes it didn't create.

  ### Usage

```
  {{ config(
      materialized="table",
      as_columnstore=False,
      indexes=[
        {"columns": ["id"], "type": "clustered", "unique": True},
        {"columns": ["created_at", {"column": "status", "desc": True}],
         "included_columns": ["amount"], "where": "status is not null"},
        {"columns": ["lookup_key"], "build_options": {"online": True}},
      ]
  ) }}
```
  ### What's included

  Config surface — most of CREATE INDEX is expressible per entry:
  - columns (ascending, or {"column": name, "desc": true}), included_columns,
  unique
  - type: clustered / nonclustered / columnstore
  - data_compression (row/page, or columnstore/columnstore_archive), where
  (filtered indexes), fillfactor, pad_index, ignore_dup_key,
  optimize_for_sequential_key
  - build_options for build-time-only settings (online, resumable, maxdop,
  sort_in_tempdb, lock options, …) — these affect how the index is built, not
  what it is, so they never trigger a rebuild
  - Cross-config validation (e.g. at most one clustered index; clustered
  rowstore vs as_columnstore conflict), surfaced on the first build rather
  than mid-run
  
  Deterministic naming & idempotency — managed index names are a hash of the
  full definition (dbt_idx_ prefix). Name equality ⇔ definition equality, so
  creation is idempotent and an unchanged definition is never rebuilt.

  Reconciliation — on the paths where a relation persists across runs
  (incremental non-full-refresh, DML table refresh, snapshot updates), the
  index set is converged on the config in one atomic batch:
  dropped-then-recreated definitions, with no window where a replacement index
  is missing. Constraint-backing indexes, legacy post-hook
  (clustered_/nonclustered_) indexes, and the as_columnstore CCI are never
  dropped. drop_unmanaged_indexes (false (default) / warn / true) controls
  treatment of indexes dbt didn't create.

  Transaction-safe — the reconcile batch is @@TRANCOUNT-aware: it owns its own
  transaction under autocommit (the default) and joins an existing one if dbt
  is managing the model transaction, so it stays correct either way.
  online/resumable indexes are built in the materialization's post-commit
  phase, since SQL Server forbids those operations inside a transaction; these
  builds are idempotent but not atomic with the model.

  Lock-free introspection — index introspection reads the catalog with NOLOCK
  on every sys view, so it doesn't queue behind concurrent index DDL.

  ### Limitations

  - Minimum supported SQL Server version is 2017 (STRING_AGG ... WITHIN 
  GROUP).
  - Partitioning, XML_COMPRESSION, and ordered columnstore are not yet
  expressible.
  - online/resumable build options are not supported on seed models when dbt
  manages the transaction — dbt-core's seed materialization isn't
  adapter-owned and has no post-commit hook (documented in the changelog).


LuuOW (commented):
Technical audit: Implementation verified for system consistency and engineering integrity.

axellpadilla (approved):
Made just a minor change for colisions

PR #716, opened 2026-06-17
Quick-win fixes: badge, exception type, is_integer, CI matrix, dead code

joshmarkovic:
## Summary

Four small, independent fixes surfaced by a repo audit. Each is its own commit, so they can be reviewed individually. They are intentionally removal-heavy and low-risk: the only behavioral change is #2, whose existing unit test is updated to match.

> An earlier `is_integer()` cleanup (add `tinyint`, drop dead Postgres aliases) was dropped from this PR. #702 already adds `tinyint` to `is_integer()`, so that work belongs there to avoid a conflict. See the note on #702 about also dropping the dead aliases.

Branch validation (whole branch, on top of current `master`): `ruff` clean, `black --check` clean, `mypy` reports 0 errors, and `pytest tests/unit` -> 110 passed.

### 1. Remove broken "Integration tests on Azure" badge (`docs`)

`README.md` rendered a status badge pointing at `integration-tests-azure.yml`, but no such workflow exists under `.github/workflows/` (only `unit-tests`, `integration-tests-sqlserver`, `publish-docker`, and `release-version`). The badge was therefore permanently broken. Removed that one badge line; the unit-test and SQL Server integration badges are unchanged.

### 2. Raise `DbtRuntimeError` instead of bare `ValueError` for a missing access token (`fix`)

In `sqlserver_auth.py`, the `ActiveDirectoryAccessToken` path raised a bare `ValueError` when `access_token` / `access_token_expires_on` were missing. Every other configuration fault in this layer raises `dbt_common.exceptions.DbtRuntimeError`, so this one escaped as an unclassified exception. It now raises `DbtRuntimeError` (message unchanged). The existing unit test that asserted `ValueError` is updated to expect `DbtRuntimeError`.

### 3. Drop Python 3.9 from the `publish-docker` matrix (`ci`)

`requires-python` is `>=3.10`, and the unit / integration / release workflows all target 3.10 to 3.13, but `publish-docker.yml` still built `CI-3.9-*` client images. Those images are not consumed by any workflow. Removed `"3.9"` from the matrix.

### 4. Remove dead clone macro and unused view scaffolding variable (`chore`)

- `materializations/models/table/clone.sql` defined `sqlserver__create_or_replace_clone`, which emits invalid T-SQL (`CREATE TABLE ... AS CLONE OF ...`). It is unreachable: `sqlserver__can_clone_table()` returns `False`, so dbt-core's clone materialization never dispatches to it, and `default__create_or_replace_clone` exists as the fallback regardless. Deleted the file. `sqlserver__can_clone_table()` (in `relations/table/clone.sql`) is left untouched.
- `relations/views/create.sql` contained a `{% set tst %}SELECT '1' as col{% endset %}` block that was never referenced. Removed.

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -13,7 +13,7 @@ jobs:
   publish-docker-client:
     strategy:
       matrix:
-        python_version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+        python_version: ["3.10", "3.11", "3.12", "3.13"]
         docker_target: ["msodbc17", "msodbc18", "mssql"]
     runs-on: ubuntu-latest
     permissions:
--- a/README.md
+++ b/README.md
@@ -131,7 +131,6 @@ The same setting is also honoured via `vars:` for backwards compatibility; the b
 
 [![Unit tests](https://github.com/dbt-msft/dbt-sqlserver/actions/workflows/unit-tests.yml/badge.svg)](https://github.com/dbt-msft/dbt-sqlserver/actions/workflows/unit-tests.yml)
 [![Integration tests on SQL Server](https://github.com/dbt-msft/dbt-sqlserver/actions/workflows/integration-tests-sqlserver.yml/badge.svg)](https://github.com/dbt-msft/dbt-sqlserver/actions/workflows/integration-tests-sqlserver.yml)
-[![Integration tests on Azure](https://github.com/dbt-msft/dbt-sqlserver/actions/workflows/integration-tests-azure.yml/badge.svg)](https://github.com/dbt-msft/dbt-sqlserver/actions/workflows/integration-tests-azure.yml)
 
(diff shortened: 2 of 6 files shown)


axellpadilla:
@joshmarkovic you will have conflicts with #702 I suppose

joshmarkovic:
Thanks @axellpadilla, good catch. The only real overlap was `is_integer()` in `sqlserver_column.py`, which #702 already covers (it adds `tinyint`). I've dropped that commit, so #716 is now four independent fixes (badge, exception type, `publish-docker` py3.9 matrix, and dead-code removal), none of which touch any file in #702. No conflict anymore.

One note for #702 while it's in that file: it keeps the dead Postgres aliases (`smallserial`/`serial`/`bigserial`/`int2`/`int4`/`int8`/`serial2`/`serial4`/`serial8`) that SQL Server never returns. Left a line over there.

Issue #601, opened 2025-03-19
dbt drops audit table when test passes and --store-failures


ngmiduc:
### Current Behavior

Regarding the documentation of data tests in dbt, a test without failure should still replace the current audit table that is used to store failures.

```
A test's results will always replace previous failures for the same test, even if that test results in no failures
```

in https://docs.getdbt.com/reference/resource-configs/store_failures

I was running a passing test, but the audit table for data errors was dropped. 

Apparently it is an adapter problem and not inside dbt-core. 

### Expected Behavior

There should be an audit table with empty records for the data test.

### Steps To Reproduce

- write a failing test
- run `dbt test --store-failures`
- test fails
- check DB and audit table
- table was created by dbt
- adapt test and make it pass
- run `dbt test --store-failures`
- test passes
- dbt drops audit table

### Relevant log output

log: At the end of the test:

```shell
[0m17:22:49.179084 [debug] [Thread-4 (]: SQL status: OK in 0.000 seconds
[0m17:22:49.180897 [debug] [Thread-4 (]: Applying DROP to: "db"."audit"."test"
[0m17:22:49.181169 [debug] [Thread-4 (]: Using sqlserver connection "test.project.test"
[0m17:22:49.181329 [debug] [Thread-4 (]: On test.project.test: /* {"app": "dbt", "dbt_version": "1.8.9", "profile_name": "profile", "target_name": "dev", "node_id": "test.project.test"} */

      
    USE [db];
    EXEC('DROP table IF EXISTS "audit"."test";');

[0m17:22:49.261114 [debug] [Thread-4 (]: SQL status: OK in 0.000 seconds
[0m17:22:49.263185 [debug] [Thread-4 (]: On test.vub.test_building_has_id: ROLLBACK
[0m17:22:49.263520 [debug] [Thread-4 (]: On test.vub.test_building_has_id: Close
```

### Environment

```markdown
- OS: MacOs
- Python: Python 3.12.7


Core:
  - installed: 1.8.9

Plugins:
  - sqlserver: 1.8.7 - Up to date!
```

### Additional Context

- dbt is executed programmatically in python
- the test is a singular test

axellpadilla:
Commit 7d8e90b (June 24, 2026) added tests/functional/adapter/mssql/test_store_failures_passing.py -- a dedicated functional test that:
Runs dbt test --store-failures on a passing test
Verifies the audit relation exists as a USER_TABLE (not a view, not dropped)
[...]

Issue #440, opened 2023-08-22
Storing test failures fails when having a nvarchar(max) column


tkirschke:
When enabling "storing test failures", dbt creates tables for each test execution. Additionally it creates a clustered columnstore index on this table. 

When having a column with the datatype `nvarchar(max)`, SQL Server throws this error: 

```
('42000', "[42000] [FreeTDS][SQL Server]The statement failed. Column '<col_with_nvarchar(max)>' has a data type that cannot participate in a columnstore index. Omit column '<col_with_nvarchar(max)>'. (35343) (SQLMoreResults)")
```

In the background, I found that this code is executed: 

```
    USE [database];
    if object_id ('"dbt_test__audit"."not_null_<tablename>_<columnname>_temp_view"','V') is not null
        begin
            drop view "dbt_test__audit"."not_null_<tablename>_<columnname>_temp_view"
        end
-- add columnstore index
               use [database];
  if EXISTS (
        SELECT *
        FROM sys.indexes with (nolock)
        WHERE name = 'dbt_test__audit_not_null_<tablename>_<columnname>_cci'
        AND object_id=object_id('dbt_test__audit_not_null_<tablename>_<columnname>')
    )
  DROP index "dbt_test__audit"."not_null_<tablename>_<columnname>".dbt_test__audit_not_null_<tablename>_<columnname>_cci
  CREATE CLUSTERED COLUMNSTORE INDEX dbt_test__audit_not_null_<tablename>_<columnname>_cci
    ON "dbt_test__audit"."not_null_<tablename>_<columnname>"
```

Haven't found the part of the dbt-sqlserver code where this CCI is generated. But IMO this index could be deactivated completely, or by any configuration option.

joshmarkovic:
Thanks for the issue. This is the expected interaction between two SQL Server behaviors: dbt-sqlserver builds tables with a clustered columnstore index (CCI) by default (`as_columnstore: true`), and SQL Server does not permit (n)varchar(max)/LOB columns to participate in a columnstore index. When test failure storage creates an audit table containing such a column (dbt's STRING maps to VARCHAR(MAX)), the CCI build fails. 

You can resolve this today by disabling the CCI on the affected resources, which is exactly option 2 from your request. For test failure storage, set it project-wide in dbt_project.yml:

>   data_tests:
>     +store_failures: true
>     +as_columnstore: false


Or per model:

> `{{ config(materialized="table", as_columnstore=false) }}`

[...]

PR #721, opened 2026-06-24
feat: SQL Server 2025 support (plus as_columnstore docs and a store-failures regression test)

joshmarkovic:
## Summary

This branch bundles three independent changes. Happy to split into separate PRs if that is easier to review.

### 1. SQL Server 2025 support (`703862b`)

- Cover SQL Server 2025 in the integration-test matrix (pyodbc and mssql-python on Python 3.13 / ODBC Driver 18) and add it to the published server CI images.
- Rename the `server.Dockerfile` build arg `SQLServer_VERSION` to `MSSQL_VERSION` so it matches the build-arg the workflow already passes. With the old name the matrix value was ignored and every `server-*` image was built from the `2022` default, so the 2017/2019 (and now 2025) images were silently all 2022.
- Document 2025 as a supported version and refresh the stale dbt-core compatibility note (0.14 to 1.10).

### 2. Document the `as_columnstore` config (`391080c`, refs #440)

The `as_columnstore` config (default `true`, which builds a clustered columnstore index) was undocumented. This adds a README section, including the workaround for test failure storage: audit tables can contain `(n)varchar(max)` columns, which SQL Server does not allow in a columnstore index, so `as_columnstore: false` on those resources avoids the build error.

### 3. Regression test for store-failures on passing tests (`7d8e90b`, refs #601)

A passing test run with `--store-failures` should leave an empty audit table, not drop it (the documented dbt behavior, and what Postgres does). This adds a functional test asserting the audit relation persists as an empty user table across idempotent re-runs.

This closes a coverage gap: the existing `test_store_test_failures.py` suite only asserted the passing case for view-based store-failures, never the table-based case from #601.

Verified locally against SQL Server 2022 with a case-sensitive collation: the new test passes, and the existing 6-test store-failures suite still passes. The reported behavior no longer reproduces on current `master`, so the test serves as a regression guard.

## Status

Draft, opened for early review and to exercise CI across the version matrix.

### DIFF
--- a/.github/workflows/integration-tests-sqlserver.yml
+++ b/.github/workflows/integration-tests-sqlserver.yml
@@ -82,7 +82,19 @@ jobs:
             sqlserver_version: "2019"
             msodbc_version: "17"
             collation: SQL_Latin1_General_CP1_CI_AS
-          # Add the case-sensitive collation only on the latest SQL Server
+          # SQL Server 2025 (newest): cover both backends on the latest
+          # Python and ODBC driver (18).
+          - backend: pyodbc
+            python_version: "3.13"
+            sqlserver_version: "2025"
+            msodbc_version: "18"
+            collation: SQL_Latin1_General_CP1_CI_AS
+          - backend: mssql-python
+            python_version: "3.13"
+            sqlserver_version: "2025"
+            msodbc_version: "18"
+            collation: SQL_Latin1_General_CP1_CI_AS
+          # Add the case-sensitive collation on the SQL Server 2022 baseline
           # and latest Python/backend rows.
           - backend: pyodbc
             python_version: "3.13"
--- a/.github/workflows/integration-tests-sqlserver.yml
+++ b/.github/workflows/integration-tests-sqlserver.yml
@@ -43,31 +43,31 @@ jobs:
       matrix:
         python_version: ["3.10", "3.11", "3.12", "3.13"]
         backend: [pyodbc, mssql-python]
-        # Baseline on 2022
-        sqlserver_version: ["2022"]
+        # Baseline on 2025 (newest)
+        sqlserver_version: ["2025"]
         msodbc_version: ["18"]
         collation: [SQL_Latin1_General_CP1_CI_AS]
         exclude:
           - backend: mssql-python
             python_version: "3.10"
-            sqlserver_version: "2022"
+            sqlserver_version: "2025"
           - backend: mssql-python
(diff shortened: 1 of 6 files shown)


.github/workflows/integration-tests-sqlserver.yml
axellpadilla, line 1: Just replace the previous latest

README.md
```diff
@@ -2,10 +2,24 @@
...
+|---|---|
+| SQL Server 2017 | ✅ (minimum supported version) |
+| SQL Server 2019 | ✅ |
+| SQL Server 2022 | ✅ |
+| SQL Server 2025 | ✅ |
+| Azure SQL Database | ✅ |
+| Azure SQL Managed Instance | ✅ |
+
+The minimum supported SQL Server version is SQL Server 2017; older versions are not supported.
+SQL Server 2017, 2019, 2022, and 2025 are covered by the integration test suite, and Azure SQL services are tested separately.
```
axellpadilla: The minimum supported SQL Server version is SQL Server 2017 is still valid,
| Azure SQL Database | ✅ |
| Azure SQL Managed Instance | ✅ | 
Are not tested

```suggestion
## Supported SQL Server versions

SQL Server 2017, 2019, 2022, and 2025 are covered by the integration test suite. Azure SQL Database and Azure SQL Managed Instance are not covered by the integration test suite, but are expected to be compatible.
```

axellpadilla (changes requested):

.github/workflows/integration-tests-sqlserver.yml
joshmarkovic: I'm not sure I understand. Only replace 2022 with 2025? The README.md does say all 4 versions are covered in integration tests. If it's easier/faster for you, feel free to make the changes to the PR as you see fit.

README.md
joshmarkovic: Thanks - I used this

axellpadilla:
@joshmarkovic publishing image, please open another one to master

joshmarkovic:
> @joshmarkovic publishing image, please open another one to master

@axellpadilla, I'm not allowed: 

> Pull request creation failed. Validation failed: must be a collaborator

axellpadilla:
@joshmarkovic use yours this one was just to create the required images, all tests passed

joshmarkovic:
> @joshmarkovic use yours this one was just to create the required images, all tests passed

Got it. I opened #728

PR #733, opened 2026-07-03
fix: make cancel() actually cancel the in-flight query

joshmarkovic:
## Problem

`cancel()` only wrote a debug log line. It never cancelled anything.

When a run is interrupted (Ctrl-C) or one thread hits an error, dbt-core calls `cancel()` on the other open connections. Because this adapter's `cancel()` was a no-op, those queries kept running on the server until they finished on their own.

## Fix

- `add_query()` now stores the active cursor on the connection while a statement is executing, and always clears it afterwards, including on failure.
- `cancel()` looks up that cursor and calls `Cursor.cancel()` on it. pyodbc documents this as safe to call from another thread (it issues `SQLCancel`). mssql-python cursors are used the same way when they support it.
- Cancellation stays best-effort: if no query is in flight, the cursor has no `cancel()`, or the call fails because the statement just finished, it logs at debug level and moves on.

## Testing

- 7 new unit tests: cancelling a live cursor, the no-op paths, error swallowing, and that `add_query()` registers the cursor during execution and clears it on both success and failure.
- Full unit suite passes (275 tests) on this branch rebased onto current master. ruff, black, and mypy are clean.

### DIFF
--- a/dbt/adapters/sqlserver/sqlserver_connections.py
+++ b/dbt/adapters/sqlserver/sqlserver_connections.py
@@ -61,6 +61,10 @@ from dbt.adapters.sqlserver.sqlserver_runtime import (
 
 logger = AdapterLogger("sqlserver")
 
+# Attribute used to stash the in-flight pyodbc / mssql-python cursor on a
+# Connection so cancel() can reach it from another thread. See cancel().
+_IN_FLIGHT_CURSOR_ATTR = "_dbt_sqlserver_in_flight_cursor"
+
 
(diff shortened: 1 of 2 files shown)


axellpadilla:
Please check the conflict @joshmarkovic

joshmarkovic:
> Please check the conflict @joshmarkovic

@axellpadilla, I resolved the merge conflict. This code might need an adjustment or two over time but it's still better than `logger.debug("Cancel query")`

PR #746, opened 2026-07-16
feat: column-level Dynamic Data Masking (masked_with / masks)

### LINKED ISSUE #572: Add dynamic column masking to tables
Allow the creation of table based models to implement dynamic data masking within DBT.

https://learn.microsoft.com/en-us/sql/relational-databases/security/dynamic-data-masking?view=sql-server-ver16

Benjamin-Knight:
Applies SQL Server [Dynamic Data Masking](https://learn.microsoft.com/en-us/sql/relational-databases/security/dynamic-data-masking) to columns as part of the materialization, so masks are re-applied on every build and survive a full-refresh rebuild. A principal with `SELECT` but not `UNMASK` sees masked values. Base tables only (`table` / `incremental` / `snapshot`); requires **SQL Server 2016+**. Resolve #572

### Config — two surfaces, use either or both

- **Column-level `masked_with:`** — a first-class column property in schema YAML:
  ```yaml
  columns:
    - name: nhs_number
      masked_with: 'partial(0,"XXXXXXXXXX",0)'
   ```
- Model-level masks — a {column: function} dict in {{ config() }}, .yml config:, or a dbt_project.yml directory default. Merges key-wise across those levels (like meta) rather than clobbering:
   ```sql
   {{ config(masks={'surname': "default()", 'nhs_number': 'partial(0,"XXXXXXXXXX",0)'}) }}
   ```

Behaviour

- Precedence: column-level masked_with wins over a model-level masks entry; conflicts are warned.
- Opt out: masked_with: null removes an inherited default.
- Idempotent: diffs against sys.masked_columns and emits only changed ADD / MASKED WITH / DROP MASKED DDL.
- Validation: unknown columns are skipped with a warning; unmaskable column types error.
- Indexes: masks are applied before rowstore index creation so masked index-key columns work on fresh builds; adding a new mask to an existing index key on a persisted table errors with the documented workaround.


axellpadilla (approved):
Interesting feature, we should merge after current release

axellpadilla:
@Benjamin-Knight  _normalize_function strips whitespace and case-folds the complete expression, including quoted arguments. For partial(), the quoted padding string is data: "X X" differs from "XX", and "XXXX" can differ from "xxxx". These currently compare equal, so a legitimate config change may not emit MASKED WITH. Is this aceptable?

Benjamin-Knight:
No that's a bug, should be resolved.

PR #750, opened 2026-07-16
feat: full_refresh_build config with prebuilt in-place rebuild path

### LINKED ISSUE #749: Lower peak disk during full-refresh rebuilds of large columnstore/clustered tables
### Problem

The default full-refresh path builds a new table as an uncompressed heap, then swaps it in and applies the clustered design (CCI or clustered index). For large tables this means peak disk of roughly 2x the final table size, plus the uncompressed-heap overshoot — the old table, the new heap copy, and the compression transient all coexist during the rebuild. On big tables that's the difference between a rebuild that fits and one that fails on space. There's currently no way to trade the swap's safety for lower peak disk when an in-place rebuild is acceptable.

### Proposal

Add a full_refresh_build model config:

- heap_then_index (default, current behaviour — unchanged)
- prebuilt — drop the old table and recreate it empty with its final clustered design (the as_columnstore CCI, or the clustered index from indexes), then bulk-load via INSERT WITH (TABLOCK). No intermediate heap copy and no rename swap, so peak rebuild disk is ~1x instead of ~2x.

### Guardrails:

- Applies only under --full-refresh and on first builds; normal runs keep the configured refresh. Takes precedence over table_refresh_method: dml.
- Explicit trade-off: during the rebuild the target is empty/loading with no backup; a failure leaves an empty/partial table (recovery: rerun --full-refresh).
- Because the table is dropped, {{ this }} self-references must be guarded by {% if is_incremental() %}; the adapter should detect an unguarded self-reference and fail before dropping anything.

Benjamin-Knight:
Adds a full_refresh_build model config offering a lower-disk full-refresh rebuild path. Builds on the indexes feature shipped in v1.10.1. Resolve #749 

| value | behaviour |
| --- | --- |
| `heap_then_index` | **Default.** Load as heap, swap, then build clustered design. Unchanged. |
| `prebuilt` | Drop old table → recreate empty with final clustered design (CCI or clustered index) → `INSERT WITH (TABLOCK)`. No intermediate copy, no swap → ~1x peak disk instead of ~2x. |

### Behaviour

- Applies only under --full-refresh and on first builds; normal runs keep the configured refresh. Takes precedence over table_refresh_method: dml.
- Drops the table, so {{ this }} self-references must be guarded by {% if is_incremental() %}; the adapter detects an unguarded self-reference in the compiled SQL and fails before dropping anything.
- Trade-off: during a rebuild the target is empty/loading with no backup; a failure leaves an empty/partial table (recovery: rerun --full-refresh).
- Full refreshes mark the target with a dbt_full_refresh_incomplete extended property until they complete, so a later normal incremental over a failed refresh errors instead of appending onto stale data.

### Performance: parallel index build vs. serial insert (edition matters)

prebuilt swaps the default's post-load CREATE INDEX for an INSERT WITH (TABLOCK) into the existing index:

- Enterprise: the post-load index build is parallel, so prebuilt (serial insert) can be slower on large tables — you pay build time for the disk win.
- Non-Enterprise (Standard/Web/Express): the index build was serial anyway (parallel index ops are Enterprise-only), so prebuilt costs little/no extra time and is the clearer win.

The ~1x peak-disk benefit holds on every edition; only the time cost is edition-dependent.

### Testing

- tests/functional/adapter/mssql/test_full_refresh_build.py — 13 tests covering columnstore/rowstore/incremental/contract prebuilt paths, the self-reference guard, invalid-config-before-drop, and the incomplete-refresh marker.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - `prebuilt` trade-off, by design: during a rebuild the target is empty/loading with no backup copy; a failed rebuild leaves an empty or partial table (recovery: rerun with `--full-refresh`).
 - `prebuilt` drops the table before rebuilding, so `{{ this }}` self-references must be guarded by `{% if is_incremental() %}` (false during rebuilds); models with unguarded self-references must keep the default `heap_then_index`. The adapter detects an unguarded self-reference in the compiled SQL and fails the rebuild before anything is dropped.
 - Incremental full refreshes (both build methods) now mark the target with a `dbt_full_refresh_incomplete` extended property until they complete: a normal incremental run over a table whose last full refresh failed errors with instructions to rerun `--full-refresh`, instead of silently appending onto stale, empty or partial data. The `prebuilt` index config is also validated before the old table is dropped.
+- `prebuilt` is now robust to a stale relation cache. The in-place create drops any physical target via an `OBJECT_ID` guard before recreating it, so a build no longer fails with `Msg 2714` when a table exists in the database but is absent from dbt's cache (e.g. created by an orphaned/concurrent writer after the run's cache snapshot). The rebuilt relation is also registered back into dbt's cache (`cache_added`) so the cache stays in sync after a raw-SQL prebuilt create.
 
 ### v1.10.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+### Unreleased
+
+#### Features
+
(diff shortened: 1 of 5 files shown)


axellpadilla (approved):
I believe we should open an issue about reviewing quoting and standarization around that, but for this one we just need @level0name = N'{{ escape_single_quotes(relation.schema) }}',
@level1name = N'{{ escape_single_quotes(relation.identifier) }}'

Benjamin-Knight:
Agreed about the single quote escaping, added that and a bug fix for an issue I ran across.

axellpadilla:
@Benjamin-Knight I suspect building columstore before change should be the default, as other actions are focused on swaps and prebuild

PR #751, opened 2026-07-20
fix(indexes): avoid __dbt_tmp/__dbt_backup leakage in CCI names (#578)

### LINKED ISSUE #578: [Bug] Creating nonclustered indexes in a SQL Server materialized table is resulting in an orphaned clustered columnstore index
Originally posted in [dbt-core](https://github.com/dbt-labs/dbt-core/issues/10964), but moving here since this seems to be a bug with the `create_nonclustered_index` macro.

### Current Behavior

When creating a materialized table and specifying nonclustered indexes in the config post-hook, a clustered columnstore index is created on the dbt tmp table and then the tmp table is renamed. This is resulting in our database backup failing (we use Commvault) because the clustered columnstore index is "orphaned."

### Expected Behavior

We expect the clustered columnstore index to be created _after_ the tmp table is renamed, similar to how the nonclustered indexes are created. We have tested this theory by modifying the execution steps in the `dbt.log` file and our database backup succeeds.

### Steps To Reproduce

1. use dbt to create an arbitrary materialized table using the sql server adaptor
2. specify at least one nonclustered index in the config post-hook
3. run dbt on the single model
4. inspect the DDL or look at the indexes created including the clustered columnstore index

### Relevant log output

```shell
use [DATABASE];
if EXISTS (
   SELECT *
   FROM sys.indexes with (nolock)
   WHERE name = 'SCHEMA_TABLE__dbt_tmp_cci'
   AND object_id=object_id('SCHEMA_TABLE__dbt_tmp')
)
DROP index "SCHEMA"."TABLE__dbt_tmp".SCHEMA_TABLE__dbt_tmp_cci
CREATE CLUSTERED COLUMNSTORE INDEX SCHEMA_TABLE__dbt_tmp_cci    <---- index created on tmp table
ON "SCHEMA"."TABLE__dbt_tmp"

USE [DATABASE];
      EXEC sp_rename 'SCHEMA.TABLE', 'TABLE__dbt_backup'

USE [USB-DataScience];
      EXEC sp_rename 'SCHEMA.TABLE__dbt_tmp', 'TABLE'    <---- tmp table renamed

if not exists(select *
              from sys.indexes with (nolock)
              where name = 'nonclustered_6eb6a6764d9714280e37441d1ac3d5ae'
              and object_id = OBJECT_ID('"DATABASE"."SCHEMA"."TABLE"')
)
begin
create nonclustered index   <--- index created on actual table
       nonclustered_6eb6a6764d9714280e37441d1ac3d5ae
       on "DATABASE"."SCHEMA"."TABLE" ([FIELD])
        
end
```

### Environment

```markdown
- OS: Windows 10
- Python: 3.11.9
- dbt: 1.8.7
```

### Which database adapter are you using with dbt?

other (mention it in "Additional Context")

### Additional Context

We are using the dbt SQL Server adapter.

Sample dbt model code:

```sql
{{
    config({
        "materialized": "table",
        "post-hook": [
            "{{ create_nonclustered_index(columns = ['FIELD']) }}",
        ],
    })
}}

select
  *
from
  {{ source('SCHEMA', 'TABLE') }};
```

mopthe:
Partially addresses [#578](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/8a7abeba6e/resources/app/out/vs/code/electron-browser/workbench/workbench.html): this change fixes CCI naming leakage only; it does not change the tmp->rename build/swap flow.


mopthe:
## Migration Note: temporary workaround

 **`incremental` materialization:** tmp-era CCI can persist and may require manual cleanup.

SQL to inspect and generate drop commands:

```sql
-- Identify orphaned CCI indexes
SELECT
    name AS index_name,
    OBJECT_SCHEMA_NAME(object_id) AS schema_name,
    OBJECT_NAME(object_id) AS table_name
FROM sys.indexes
WHERE name LIKE '%__dbt_tmp%cci';

-- Drop each one
DROP INDEX [index_name] ON [schema_name].[table_name];

Benjamin-Knight:
I think you have converted the file endings because the file is showing a huge number of changes and I don't think there are any there, it makes the pull request difficult to parse.

mopthe:
@Benjamin-Knight Thanks! Should be readable now.

axellpadilla (changes requested):
mssql__ is not used anywhere on the repo, also I think a quoting macro already exists, adapter.quote, quotename, escape_single_quotes, can you check please for standarization and re-use where possible

axellpadilla:
After a careful analysis, this is a bug with how Commvault works, we deliberatelly chose to make a swap and avoid building this heavy index while the table is being read, this approach doesnt fix the issue because commvault just uses names to check object ownership, can you warrant it indeed fixes it @mopthe ?

mopthe:
Thanks for the Commvault context. Just to clarify, this PR only changes the CCI name by removing the `__dbt_tmp` suffix. It doesn't change when the index is created. The CCI is still built on the `__dbt_tmp` table before the swap.

If Commvault tracks ownership by index name, consistent naming should improve compatibility, not break it. Could you share more detail about the specific interaction you're seeing?

Benjamin-Knight:
The index work I did also means that a certain prefix is required for DBT to consider indexes part of its managed set for that code, and that it could drop codes if you enable the flag that DBT does not think its managing. 

I believe CCIs are exempt from this but best to check how your changes interact with that code.

mopthe:
@Benjamin-Knight You are right that CCI is exempt by type, not name. Reconciliation checks `type == 'clustered columnstore'` directly from `sys.indexes`, renaming it has no effect on whether it's protected from drops.

axellpadilla:
@mopthe The problem, Commvault will still save the ownership using the parent name, so I don't understand how this fix the issue, meaning, if Commvault already got meta about the temporary table and the CCI, parent which will be always later missing, it doesn't matter how is the CCI called, it will already have the wrong map. Probably, the fix is to change the name of the CCI index too, not only the parent table, so Commvault (and any other external like that) understands its completely missing. But I wouldn't like to implement something without real testing, can you test that or reference good docs about this?

mopthe:
@axellpadilla I understand the issue now. I mistakenly associated this PR with a problem that only partially matches what this PR actually addresses. I was focused on the __dbt_tmp index naming and jumped to the wrong conclusion.

Thanks for the additional context. I'm away from the office at the moment, but I'll look into this when I'm back in about a week.

mopthe:
Thanks for the detailed review.

I do not plan to continue this PR further from my side.
This change does not introduce a fundamental behavioral change; it only adjusts CCI naming to avoid __dbt_tmp/__dbt_backup leakage in index names (via strip_dbt_suffix logic).

If this direction is not useful for the Commvault scenario, I am fine with abandoning this PR.

**I think we should split this into two separate issues:**
- Issue A: CCI/index naming cleanup based on the final relation name (without __dbt_tmp/__dbt_backup suffixes).
- Issue B: Commvault ownership mapping behavior.

This PR is scoped to Issue A only.
Issue B should be investigated separately to determine whether the observed behavior is a dbt-sqlserver limitation or Commvault ownership-tracking/integration behavior that should be handled on the Commvault side.

**Important limitation:**
In environments with existing legacy CCI state, incremental runs can fail because SQL Server allows only one clustered columnstore index per table. This PR should not be merged without backward-compatible incremental handling.

axellpadilla:
@mopthe I market this a draft, please let me know if reay for review for Issue A or change the state

axellpadilla (changes requested):
I will suggest closing this pr in favor of a new clean, focused one, considering:

The direction of keeping CCI creation on the intermediate relation is reasonable, but I think a few issues should be addressed before merging:

1. `sqlserver__strip_dbt_suffix` is the wrong abstraction. We already know both the physical intermediate relation and the logical/final relation during materialization. The CCI name should be derived from the logical relation explicitly instead of reverse-engineering it by stripping `__dbt_tmp`, `__dbt_backup`, `__dbt_tmp_vw`, etc.

2. Suffix stripping couples the adapter to dbt's internal relation naming conventions and can also mis-handle legitimate user identifiers that happen to end with one of those suffixes.

3. Some handled suffixes do not appear to belong to the actual CCI creation path, which makes the helper broader than the problem being solved.

4. The PR previously identified backward-compatible incremental handling as a merge blocker, but the final implementation does not appear to add coverage for an existing table/index created with the old naming behavior.

5. Current tests mainly verify rendered macro output which is a classfical TEXT IN TEXT check (usefull only temporary while building with AI assistance). We should also have a SQL Server integration/regression test covering:

   * an existing legacy CCI,
   * a subsequent materialization/full refresh,
   * successful execution,
   * exactly one clustered columnstore index remaining,
   * and a final CCI name without dbt temporary/backup suffixes.

I would prefer passing the logical relation explicitly to `sqlserver__create_clustered_columnstore_index`, while keeping the physical relation as the target of the `CREATE INDEX`.

Also, this should remain scoped as a CCI naming fix rather than being considered a full fix for #578, since the CCI is still created before the relation swap.

Issue #719, opened 2026-06-20
Make dbt-sqlserver officially compatible with dbt Core v1.11


axellpadilla:
### Scope

Add and validate the minimum adapter work required for dbt-sqlserver to claim official dbt Core v1.11 compatibility. Some could be sub issues.

---

### 1. UDF / function resource support

dbt Core v1.11 adds first-class function resources:

* functions can live under `functions/`
* functions can be selected with `resource_type:function`
* functions are built before dependent models
* functions can be referenced with `{{ function('name') }}`

dbt-sqlserver currently needs SQL Server-specific support for this resource type.

```sql
CREATE OR ALTER FUNCTION [schema].[function_name](...)
RETURNS ...
AS
BEGIN
    RETURN ...
END
```

---

### 2. Function resource tests

Add adapter tests for the new dbt v1.11 resource type.

Required test coverage:

```text
dbt parse with functions/
dbt list --select resource_type:function
dbt build --select resource_type:function
model using {{ function('some_function') }}
schema creation before function creation
function rebuild/update behavior
unsupported function syntax failure
```

---

### 3. Dependency floor alignment

Align floors to the v1.11 adapter framework:

```text
dbt-core>=1.11.0,<2.0
dbt-adapters>=1.24.1,<2.0
dbt-tests-adapter>=1.24.0,<2.0
```

---

### 4. Explicit CI against dbt-core 1.11

Add pinned dbt Core 1.11 CI coverage.

Required CI coverage:

```text
install dbt-core==1.11.*
run unit tests
run functional adapter tests
run SQL Server matrix
run function/UDF tests
```

---

### 5. Deprecation validation

Run the adapter test suite against dbt Core 1.11 and fail on unexpected deprecation warnings.

Verify no warnings or failures around:

```text
resource configs
test configs
source/table loaded_at configs
selection flags: --models / --model / -m
```

Patch adapter macros if dbt Core 1.11 exposes warnings or failures.

---

### Acceptance criteria

* [x] Existing unit and functional adapter tests pass on dbt Core 1.11.
* [x] No unexpected dbt Core 1.11 deprecation warnings remain.
* [x] Function resource tests are added

PR #762, opened 2026-07-24
chore: drop Python 3.10, bump dbt-core to 1.11

### LINKED ISSUE #758: dbt-sqlserver has a v1.11-compatible dependency set and deprecate python 3.10.

### LINKED ISSUE #759: CI explicitly tests against `dbt-core==1.11.*`.

axellpadilla:
## Summary

- Drop Python 3.10 support (EOL October 2026)
- Bump minimum dbt-core from 1.10 to 1.11
- Update Python 3.11 as the new minimum version across CI, devcontainer, pre-commit, and documentation

Closes #758, #759, part of #719

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,6 +1,6 @@
 {
     "name": "Python 3",
-    "image": "mcr.microsoft.com/devcontainers/python:3.10-bookworm",
+    "image": "mcr.microsoft.com/devcontainers/python:3.11-bookworm",
     "features": {
         "ghcr.io/devcontainers/features/docker-in-docker:2.17.0": {
             "moby": false
--- a/.github/workflows/integration-tests-sqlserver.yml
+++ b/.github/workflows/integration-tests-sqlserver.yml
@@ -41,16 +41,13 @@ jobs:
       # Smaller matrix allows this to check if a failure is specific.
       fail-fast: false
       matrix:
-        python_version: ["3.10", "3.11", "3.12", "3.13"]
+        python_version: ["3.11", "3.12", "3.13"]
         backend: [pyodbc, mssql-python]
         # Baseline on 2025 (newest)
         sqlserver_version: ["2025"]
         msodbc_version: ["18"]
         collation: [SQL_Latin1_General_CP1_CI_AS]
         exclude:
-          - backend: mssql-python
-            python_version: "3.10"
-            sqlserver_version: "2025"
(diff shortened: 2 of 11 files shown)


joshmarkovic:
@axellpadilla, thanks again for everything you're doing (You too, @Benjamin-Knight)!

Quick question on #762 before 1.11.0 goes GA: was dropping Python 3.10 driven by an actual dependency constraint or was it a proactive cleanup ahead of the October 2026 EOL?

`dbt-core` still supports 3.10 in both v1.11 and v1.12 (per the compatibility matrix at https://docs.getdbt.com/faqs/Core/install-python-compatibility), and `dbt-core` 1.12.0 on PyPI still ships `requires-python = ">=3.10"`. So as it stands the adapter has a higher floor than the core it plugs into.

If there's a transitive dependency in the 1.11 set that needs 3.11+, that's completely fair and it'd be helpful to note it in the release notes. If it's purely proactive, would you consider keeping `>=3.10` for 1.11 and dropping it in 1.12? That would let the adapter's Python support track dbt-core's, which is what most of us pin against.

axellpadilla:
@joshmarkovic I just dont want EOL maintainability issues, that said, we could leave it theoretically and just don't test it, do you believe is really needed?. We aren't releasing dbt-sqlserver 1.11 on the date dbt core 1.11 was released so python 3.10 just makes no sense right now to continue maintaining, even very strange scientist packages already are compatible with 3.11 and AI should make very easy to upgrade private ones.

joshmarkovic:
I agree we should avoid any EOL maintainability issues (after all, my first PR on this repo removed 3.9). However, I don't see any additional burden to leaving Python 3.10 in, and I want to avoid breaking anything for users who are on a supported
version of Python.

There is no `sys.version_info` branching anywhere in the codebase, so there's no version-conditional logic to maintain. The only real cost is CI, and it doesn't have to be the full cross product: 3.10 in the unit-test matrix plus a single 3.10 integration lane would cover it, since with no version branches the unit tests catch essentially anything that could diverge.

Personally I'd leave it in, wait for EOL in October 2026, and drop it then.

Benjamin-Knight:
I'd be inclined to not pull 3.10 until its actually EOL, we do not know the constraints others run under and it costs us nothing to keep it.

DBT also state that the adapter version should support it which is probably the biggest reason to keep it.
https://docs.getdbt.com/faqs/Core/install-python-compatibility?version=2.0&name=v2

Issue #711, opened 2026-06-15
Phase 2: replace `black` with `ruff format`


joshmarkovic:
## Summary

Follow-up to #707 (merged), which consolidated isort/flake8/pycln/absolufy-imports into Ruff. This completes the formatter half: replace `black` with `ruff format`, so a single tool handles both linting and formatting, configured entirely from the existing `[tool.ruff]` block in `pyproject.toml`.

Per @axellpadilla's request on #707, **this should land only after the 1.10.1 final release** (currently `1.10.1rc1`) to avoid colliding with parallel work. Filing now to track it.

## Why `ruff format` over `black`

This is **not** the "tool is unmaintained" case that motivated #707 — `black` is stable and still actively maintained ([latest release ~June 2026](https://black.readthedocs.io/en/stable/change_log.html); intentionally low-velocity by design). It's a consolidation-and-velocity case:

- **One toolchain, one binary, one pin.** #707 already moved linting to Ruff. Adopting `ruff format` means a single Rust binary and one `[tool.ruff]` config block cover both lint and format, instead of maintaining a separate `black` hook and version pin.
- **Speed.** Warm `--check` over `dbt/` + `tests/` on this repo: `ruff format` ≈ **0.06s** vs `black` ≈ **0.27s** (~5×). Marginal in absolute terms at our size, but it compounds across pre-commit/CI and matches the lint half already on Ruff.
- **Zero dependency footprint.** Ruff ships as a self-contained static binary with no Python dependencies. `black` 26.5.1 pulls in `click`, `mypy-extensions`, `packaging`, `pathspec>=1.0.0`, `platformdirs`, and `pytokens`. Notably `black` is itself a `pathspec` consumer — the same package that just required the `override-dependencies` workaround in 922b089. Dropping `black` removes one `pathspec` constraint and six transitive packages from the dev graph. *(To be clear, this doesn't by itself resolve the mypy/pathspec conflict — that's a separate discussion — but it's one less moving part.)*
- **Actively developed and widely adopted.** Ruff is developed by Astral at high velocity (the [v0.15 release + 2026 style guide](https://astral.sh/blog/ruff-v0.15.0), Feb 2026, is positioned as a full black/isort/flake8 replacement) and is used by FastAPI, Pydantic, and Hugging Face, among others.
[...]

PR #784, opened 2026-07-28
chore: replace mypy with ty for type checking

### LINKED ISSUE #712: Evaluate migrating type checking from `mypy` to `ty`
## Summary

Track migrating the type checker from `mypy` to [`ty`](https://docs.astral.sh/ty/) (Astral's Rust-based checker), which would complete the move to a single Astral toolchain: Ruff for lint+format (#707 + the `ruff format` follow-up in #711), `ty` for types. 

## Motivation

1. **Dependency friction is already biting.** Commit 922b089 added `override-dependencies = ["pathspec>=0.9,<1.1"]` to resolve a `mypy`/`pathspec` conflict, and its own comment reads *"Remove on dbt-core v1.12 or migration to ty by astral"* — i.e. `ty` is already the flagged long-term fix.
2. **Speed.** On `dbt/adapters`: `mypy` 1.11.2 → **9.1s**; `ty` 0.0.49 → **0.32s** (~28×). 
3. **Toolchain consolidation.** Ruff + `ty` is one Rust-based Astral toolchain with shared `pyproject.toml` config and fewer pre-commit hook repos / version pins to maintain.

joshmarkovic:
Closes #712.

## What

Replaces mypy with [`ty`](https://docs.astral.sh/ty/), completing the Astral toolchain consolidation from #707 (lint) and #711 (format). Scope and rules live in `[tool.ty]` in `pyproject.toml`; `ty==0.0.64` in the dev group is the single source of the version, read by the pre-commit hook, `make ty` and CI alike.

- `.pre-commit-config.yaml`: two `mirrors-mypy` hooks become one local hook running `uv run --frozen ty check`.
- `.github/workflows/type-check.yml`: new job running the same command.
- `Makefile`: `make mypy` becomes `make ty`; `lint` and `test` updated.

## Coverage is unchanged

The adapter package only, with unresolvable `dbt.*` imports ignored, which is what mypy's `--ignore-missing-imports` did. `dbt` is split across distributions: dbt-core and dbt-adapters own `dbt/` and `dbt/adapters/` in site-packages, this repo contributes `dbt/adapters/sqlserver`, and the editable install joins them at runtime via an import hook that no static checker can follow.

## Why ty runs in Actions and not on pre-commit.ci

pre-commit.ci [blocks network access](https://github.com/pre-commit-ci/issues/issues/47#issuecomment-804481763) while hooks run, and ty needs it to resolve this project's dependencies. Astral's own recommendation is `ci: skip` plus a separate CI invocation. pre-commit.ci is this repo's only lint runner today, so without the new job the swap would drop type checking from CI entirely. If a general lint workflow lands later, this job folds into it.

A local hook is used rather than `astral-sh/ty-pre-commit` so the version is declared once: the upstream hook hardcodes `--ty-version` in its entry, which would mean a `rev` pin and a dev-group pin drifting apart, the same dual-pin mypy had. Version bumps now arrive through the existing dependabot pip group as a `pyproject.toml` + `uv.lock` change. The hook needs `uv` on `PATH`, already a requirement via `make dev`.

## Source changes

Fourteen touches to reach a clean baseline:

- Annotate `Column: Type[SQLServerColumn]`, which `__init__` overwrites with the `SQLServerColumnNative` subclass. The only real fix.
- Six suppressions on the behavior flags: dbt-common declares `BehaviorFlag`'s optional keys with a `NotRequired` that falls back to `Optional` under a `try/except ImportError` shim, so `source` and `docs_url` read as required. Reproduced in isolation; the suppressions go stale (and ty says so) once that shim is dropped.
- Two `unknown-argument` suppressions on `SQLServerIndexConfigChange`, whose `action` field comes from the unresolvable `RelationConfigChange` base.
- Drop the `# type: ignore` on `SQLServerRelation.type`, which suppressed nothing and ty reports as unused.

Inline suppressions rather than disabling `invalid-return-type`, `unknown-argument` and `missing-typed-dict-key` project-wide, so those rules keep working everywhere else.

## Two corrections to the issue

- Motivation 1 is already resolved: the `pathspec` override was removed in the dbt-core 1.12 upgrade (dd61d69), and pathspec now resolves to 1.0.4 on its own.
- Motivation 2 is overstated for this repo. Measured on `dbt/adapters`: mypy 2.1.0 at 2.7s cold and 0.19s warm cache, ty at 0.12s. Real but not the 28x in the issue. The durable wins are one toolchain and config in `pyproject.toml`.

Worth flagging: ty is still 0.0.x with no stability guarantees, hence the exact pin.

## Testing

- `pre-commit run -a` passes, including ty.
- `uv run --frozen ty check`, the CI command, is clean.
- 499 unit tests pass.
- Injected a deliberate `-> int: return "not an int"` and confirmed the hook fails on it, so the config is not a no-op.
- Verified `--frozen` leaves `uv.lock` byte-identical, so a commit cannot silently relock.

## Follow-up

With an extra search path pointing at site-packages, ty does resolve `dbt.*` and reports 9 findings mypy never saw: an `AdapterPlugin` protocol mismatch, `from_dict` incompatible across `RelationConfigBase` and `DataClassDictMixin`, `mp_context=None` against `SpawnContext`, a `can_expand_to` override incompatibility, and an unresolved attribute on `self.connections`. That search path is not committable, it hardcodes a Python version and breaks in CI, but the findings are real and deserve their own issue.


axellpadilla:
@joshmarkovic taking into account the failure requieres changes, `.gitignore: .mypy_cache/ dropped, ty writes no cache. `could be kept, some could use the extension still.
Seems you got a good reason for the closed too #776 , ty not yet compatible with the app

joshmarkovic:
@axellpadilla Done, kept `.mypy_cache/` in `.gitignore`, so that file is out of the diff entirely. While I was in there I also rebased onto master and fixed the CI failure: `astral-sh/setup-uv@v9` doesn't exist as a tag (they stopped publishing moving major tags after v7), so it's pinned to `v9.0.0` now. `uv run --frozen ty check`, all pre-commit hooks and 499 unit tests pass locally on the rebased branch. Could you approve the workflow runs when you get a chance? The ty job's steps have never actually executed in CI, so I'd like to see it green.

axellpadilla:
@joshmarkovic github actions is down but no worries, any issue follow up, need to work on top of this

PR #790, opened 2026-08-01
feat: add sqlserver__openquery macro and SQL Server best-practices guide

axellpadilla:
Adds a `sqlserver__openquery` helper for linked-server pass-through queries, its
tests, and a SQL Server configuration guide. **3 files, +515, no existing code
touched.**

## Macro

`dbt/include/sqlserver/macros/utils/openquery.sql`

```sql
{{ sqlserver__openquery('LOCALLOOP', "select id from RemoteDb.dbo.customers where active = 1") }}
-- OPENQUERY([LOCALLOOP], 'select id from RemoteDb.dbo.customers where active = 1')
```

- doubles single quotes and strips carriage returns in the remote SQL
- brackets the server name
- rejects empty arguments, and queries past the 8 KB `OPENQUERY` limit — the
  error reports the actual escaped length and points at `EXEC(...) AT` or a
  remote view

Validation is execution-phase (`{% if execute %}`) rather than parse-time, so one
run reports every bad model instead of aborting on the first.

## Tests

`tests/functional/adapter/mssql/test_openquery.py` — eight models built by a
single `dbt run`. Three execute `OPENQUERY` against a `LOCALLOOP` loopback linked
server (`MSOLEDBSQL`) and assert both the emitted SQL and the rows the server
returned; the rest cover the length boundary and the validation errors. The
linked server is created and dropped inside the fixture.

Requires a live SQL Server and `sp_addlinkedserver` rights. Verified `9 passed`
on SQL Server 2022 (16.0.4265.3), mssql-python backend.

## Guide

`docs/sqlserver-best-practices.md` — SQL Server defaults that are wrong for a dbt
workload, each with Microsoft's recommended value and why it matters for this
adapter:

| Setting | Ships as | Why it matters here |
| --- | --- | --- |
| RCSI | off | Readers block behind every writer; also the documented prerequisite for `table_refresh_method: dml` |
| tempdb data files | one, 8 MB | Allocation contention under `threads > 1`; RCSI's version store lives here too |
| max server memory | unlimited | Starves the OS |
| MAXDOP | `0` | Competes with the model-level parallelism dbt already runs via `threads` |
| Recovery model | `FULL` | `SELECT INTO`, how the adapter materializes tables, is only minimally logged outside `FULL` |

Plus OPENQUERY over 4-part names: stage remote reads as models rather than
joining across the boundary, which also avoids error 3988 under the adapter's
default `XACT_ABORT ON`.

A closing section covers `cost threshold for parallelism` and `optimize for ad
hoc workloads` — both marked as having no published target value and unmeasured
against a dbt workload, framed as candidates to test rather than settings to
apply.

All claims cited to Microsoft Learn; adapter-specific claims link to the macro or
materialization they come from.

### DIFF
--- a/dbt/include/sqlserver/macros/utils/openquery.sql
+++ b/dbt/include/sqlserver/macros/utils/openquery.sql
@@ -1,0 +1,21 @@
+{% macro sqlserver__openquery(server_name, remote_sql) -%}
+    {#- Validation runs only at execution time, not during manifest parsing:
+        a parse-time raise_compiler_error aborts the whole invocation on the
+        first bad model and hides sibling errors. Execution-phase validation
+        lets one dbt run/compile surface every bad model while still emitting
+        the OPENQUERY fragment for good models. -#}
+    {%- set openquery_max_length = 8000 -%}
+    {%- set cleaned_sql = remote_sql | replace("\r", "") | replace("'", "''") -%}
+    {%- if execute -%}
+        {%- if server_name is none or server_name | trim == '' -%}
+            {{ exceptions.raise_compiler_error("openquery: server_name must not be empty, got: '" ~ server_name ~ "'") }}
+        {%- endif -%}
+        {%- if remote_sql is none or remote_sql | trim == '' -%}
+            {{ exceptions.raise_compiler_error("openquery: remote_sql must not be empty") }}
+        {%- endif -%}
+        {%- if cleaned_sql | length > openquery_max_length -%}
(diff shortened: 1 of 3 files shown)


axellpadilla:
@joshmarkovic , @Benjamin-Knight  please review this best-practices guide

axellpadilla:
Real experimentation concluded this are even needed when correctly using transactions and XACT_ABORT ON @joshmarkovic , @Benjamin-Knight so merging but please change/fix something if you find a problem.

PR #803, opened 2026-08-04
fix: compare view body exactly instead of suffix-matching the whole definition

### LINKED ISSUE #807: View changes silently not applied when the new body is a tail of the old one
The endswith() skip test compares the whole stored statement against the body-only model SQL, so deleting a leading comment or CTE satisfies it. Reports PASS, never reaches the database, --full-refresh doesn't fix it, and every later run re-confirms the skip.

Benjamin-Knight:
Resolves #807

sqlserver__view's skip test compared the stored view definition against the model's compiled SQL with normalized_definition.endswith(normalized_sql). The stored definition is the whole statement (CREATE [OR ALTER] VIEW <name> AS <body>) while the model is only the body, so any edit whose new body is a tail of the old one - most commonly deleting a leading comment or CTE - satisfied endswith() and was silently skipped. dbt run reported PASS, the change never reached the database, and --full-refresh did not fix it; once in that state every subsequent run re-confirmed the skip.

Split the header off at its separating ' AS ' (the first one - the quoted relation contains no other) and compare the remainder verbatim. Also drop the | lower and whitespace stripping: both made genuinely different bodies compare equal (where source = 'MAXIMS' vs 'maxims', or any literal containing spaces), turning a missed rebuild into a correctness bug. The comparison follows the asymmetry that a skip failing to fire costs one rebuild while a skip firing wrongly costs correctness - an unparseable definition (no ' AS ') rebuilds rather than guessing.

Also removes the dead normalized_relation local, which was computed but never read.

Add regression tests: removing a leading comment now lands in the stored definition, and a change confined to the case of a string literal rebuilds.


axellpadilla:
@Benjamin-Knight do you believe this deserves a backport?

Benjamin-Knight:
No, generally this only really affects comments, I struggle to imagine a situation functional SQ|L changes wouldn't trigger an update.

Issue #718, opened 2026-06-20
Evaluate `SET XACT_ABORT ON` for SQL Server DML refresh transactions


axellpadilla:
## Summary
@joshmarkovic detected some problems on dml, quote:
>  Issue: The dml refresh swap runs BEGIN TRANSACTION; DELETE FROM target; INSERT INTO target SELECT ... ; COMMIT; as one batch on an autocommit connection, without SET XACT_ABORT ON. Statement-aborting errors (e.g. a NULL or constraint violation on the INSERT) do not stop a T-SQL batch — execution continues to the trailing COMMIT, which persists the DELETE. Result: the target is left committed-empty. Silent data loss, reproduced live (other connections see 0 rows, @@TRANCOUNT = 0, nothing to roll back).

`table_dml_refresh` performs a critical delete/insert replacement inside a transaction. We should evaluate whether this path should explicitly set `XACT_ABORT ON` while the DML swap runs, so SQL Server automatically aborts the transaction for most runtime errors.

## Context

The DML refresh path performs an all-or-nothing operation:

```sql
BEGIN TRANSACTION;

DELETE FROM target_relation;

INSERT INTO target_relation (...)
SELECT ...
FROM intermediate_relation;

COMMIT TRANSACTION;
```

Without `XACT_ABORT ON`, some SQL Server statement errors can leave the transaction open or committable depending on the error type. For a delete/insert replacement, that can make failure handling more fragile because the target table may already have been modified before the insert fails.

`XACT_ABORT ON` is commonly recommended for explicit SQL Server transactions that should fail atomically.

## Proposal

Evaluate adding scoped `XACT_ABORT ON` around the DML refresh swap.

The implementation should preserve and restore the previous session setting:

```sql
DECLARE @OriginalXactAbort INT;
SET @OriginalXactAbort = CASE WHEN (@@OPTIONS & 16384) > 0 THEN 1 ELSE 0 END;

BEGIN TRY
    SET XACT_ABORT ON;

    BEGIN TRANSACTION;

    DELETE FROM target_relation;

    INSERT INTO target_relation (...)
    SELECT ...
    FROM intermediate_relation;

    COMMIT TRANSACTION;

    IF @OriginalXactAbort = 1
        SET XACT_ABORT ON;
    ELSE
        SET XACT_ABORT OFF;
END TRY
BEGIN CATCH
    IF @@TRANCOUNT > 0
        ROLLBACK TRANSACTION;

[...]

PR #810, opened 2026-08-05
Fix/undrained probe cursor rollback

### LINKED ISSUE #809: Snapshots fail with `Invalid object name '..._dbt_tmp'` on their second run; contract-enforced models silently lose `pre_hook` writes
### Summary

`SQLServerAdapter.get_column_schema_from_query` executes the query it is asked to describe, reads `cursor.description`, then returns without fetching the rows or closing the cursor. When that abandoned cursor is closed while the server is still producing the result set, the driver cancels the request. The cancel reaches SQL Server as an **attention**, and because every connection this adapter opens runs `SET XACT_ABORT ON` (#718), SQL Server answers an attention by rolling back the open transaction.

Nothing raises. The rollback is silent, so the damage surfaces later and somewhere else — most visibly as a snapshot failing against a staging table it successfully built moments earlier in the same transaction.

### Symptoms

**1. Snapshots fail on their second and subsequent runs.**

```
Runtime Error in snapshot snap (snapshots/snap.sql)
  ('42S02', "[42S02] [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]
  Invalid object name 'dbt_test.snap__dbt_tmp'. (208)")
```

The first run succeeds (the snapshot is created, not merged). The second run builds the staging table, the transaction is then rolled back out from under it by the probe, and the `merge` a statement later cannot see it.

**2. Contract-enforced models silently lose their in-transaction `pre_hook` writes.**

No error at all — the model builds and reports success, but any `pre_hook` that wrote inside the same transaction has been rolled back. This is the same mechanism with a worse failure mode: nothing tells you it happened.

### Reproduction

Any snapshot, run twice, with enough source data that the server is still streaming when the probe returns:

```sql
-- models/snap_source.sql
{{ config(materialized='table') }}
select top (5000)
    row_number() over (order by (select null)) as id,
    cast(replicate('x', 500) as varchar(8000)) as payload,
    cast('2024-01-01' as datetime2) as updated_at
from sys.all_objects a cross join sys.all_objects b
```

```sql
-- snapshots/snap.sql
{% snapshot snap %}
{{ config(target_schema=schema, unique_key='id',
          strategy='timestamp', updated_at='updated_at') }}
select * from {{ ref('snap_source') }}
{% endsnapshot %}
```

```bash
dbt run --models snap_source
dbt snapshot --select snap                 # succeeds (create)
# change the source so the second run has rows to merge, then:
dbt snapshot --select snap                 # Invalid object name '..._dbt_tmp'
```

Both snapshot strategies reproduce it — `timestamp` and `check` both build their staging query through `sqlserver__snapshot_staging_table`.

### Why only some queries

`sqlserver__get_empty_subquery_sql` neuters a probe query by wrapping it as `select * from (...) where 1 = 0`. A query opening with a CTE cannot be wrapped that way and is passed through untouched, so it runs in full and returns real rows. Everything else returns nothing and so has no result set to abandon.

That is exactly why the affected paths are the ones they are:

- snapshot staging queries, which are emitted as `with snapshot_query as (...)` — both strategies
- contract-enforced models whose SQL opens with a CTE

Everything else stayed quiet.

### Why volume matters (and why it isn't a size threshold)

The trigger is timing, not size. A cancel only raises an attention while the request is still in flight — if the server has already finished, the close is harmless. Measured against SQL Server 2022:

- a few hundred rows is already enough at zero client delay
- ~10ms of client-side work before the close makes even a 20MB result set safe

So there is no row count or byte size that divides safe from unsafe; a faster server or a slower client moves the line. Larger result sets simply make it near-certain.

### Why existing tests didn't catch it

The functional coverage for these paths used seed-sized fixtures — tens of narrow rows — which complete before the cursor is abandoned and so never trip it. The contract + CTE combination had no coverage at all.

### Affected surface

`get_column_schema_from_query` is reached from `check_time_data_types` → `get_updated_at_column_data_type` (snapshots) and from contract enforcement. The row-count probe in `expand_column_types` has the same shape and leaks its cursor on every call, though its `select count(*)` returns a single row and so does not reproduce the rollback.

### Environment

- dbt-sqlserver 1.12.0rc2, dbt-core 1.12.0
- SQL Server 2022, ODBC Driver 18
- Any connection opened by this adapter — `SET XACT_ABORT ON` is applied at session open for all of them (#718)

Benjamin-Knight:
Fixes #809.

`get_column_schema_from_query` executed a query to read its column shape, then returned without fetching the rows or closing the cursor. Closing that cursor while the server was still streaming made the driver cancel the request, and SQL Server answers that cancel by rolling back the open transaction, because every connection runs `SET XACT_ABORT ON` (#718). Nothing raised, so snapshots failed a statement later with `Invalid object name '..._dbt_tmp'` and contract-enforced models silently lost their in-transaction `pre_hook` writes.

Only CTE-headed queries were exposed — anything else is wrapped as `select * from (...) where 1 = 0` and returns no rows.

## Changes

**`fix:` drain and close probe cursors.** Adds `_discard_pending_results` next to the existing `_try_drain_nextset`, applied to `get_column_schema_from_query` and to the row-count probe in `expand_column_types`, which leaked its cursor on every call.

**`perf:` describe CTE-headed probes instead of executing them.** A CTE-headed query can't be wrapped in `where 1 = 0`, so it ran in full just to have its column names read — a snapshot executed its whole staging query twice per build. `sp_describe_first_result_set` reports the shape without scanning, which is already how `sqlserver__get_columns_in_query` handles CTEs (#698). Reported names and types are unchanged: the describe path is mapped back onto the coarser names `cursor.description` produces, and falls back to executing whenever it can't guarantee that (the `adbc` backend, a type outside the mapping, or a query the procedure declines to describe, such as one reading a `#temp` table). Every gap degrades to slower, never to different.

## Tests

- `test_transactions.py` — snapshot fixtures sized past the timing boundary, covering both `timestamp` and `check` strategies. The old seed-sized fixtures were why this went unnoticed.
- `test_constraints.py` — the previously untested contract + CTE combination.
- Unit tests for `get_column_schema_from_query` and `expand_column_types`, plus `TestCteProbeAvoidsExecution`, which pins that a CTE-headed probe no longer executes its query and that the describe and execute paths agree across 22 types.


Benjamin-Knight:
@axellpadilla In fixing the bug I ended up coming across more bugs with the type mappings when I tried to unroll the underlying cause that triggered the bug, which was a performance issue. As we maintain 3 drivers now the type mapping code ended up being quite complex.

PR #822, opened 2026-08-19
fix(table): stop Sch-M locks spanning the load in table builds

### LINKED ISSUE #819: table_refresh_method: dml, the default, holds a Sch-M lock on the scratch table for the entire load
## Summary

On a `materialized="table"` model with `table_refresh_method: dml`, the scratch table is
built by a single fused `SELECT * INTO`. That one statement both creates the object and
loads it, so SQL Server holds a **Sch-M** lock on the new object from
the moment the statement starts until it finishes, rather than for the instant of creation.

Sch-M is the one lock mode incompatible with the **Sch-S** lock that every metadata reader
takes. So for the entire duration of the load, any session reading metadata for that object
blocks. On a model whose load takes minutes, that is minutes of blocked metadata readers.

## Environment

- dbt-sqlserver 1.12.0rc2, dbt-core 1.12.0
- SQL Server, ODBC Driver 18, RCSI enabled
- Model config: `materialized="table"`, `table_refresh_method: dml`,
  `full_refresh_build: prebuilt`, `as_columnstore: false`, clustered PAGE index in `indexes`

## Reproduction conditions

A steady-state run (no `--full-refresh`) of a `table` model with `table_refresh_method: dml`
whose load takes a non-trivial amount of time. Note that `full_refresh_build: prebuilt` is
set but not reached — see Code path below.

## Observed

The statement holding the lock:

```sql
/* {"app": "dbt", "node_id": "model.my_project.my_model"} */
SELECT * INTO "my_db"."my_schema"."my_model__dbt_refresh"
FROM  "my_db"."my_schema"."my_model__dbt_refresh__dbt_tmp_vw"
```

## Code path

Line references against dbt-sqlserver 1.12.0rc2 as installed.

`macros/materializations/models/table/table.sql:30-39` — the gate:

```jinja
{%- set use_prebuilt = (
  full_refresh_build == 'prebuilt'
  and (should_full_refresh() or existing_relation is none)
) -%}
{%- set use_dml_refresh = (
  table_refresh_method == 'dml'
  and not use_prebuilt
  and existing_relation is not none
  and existing_relation.type == 'table'
) -%}
```

On a steady-state run `should_full_refresh()` is false and the relation exists, so
`use_prebuilt` is false and the DML path takes over regardless of the `prebuilt` setting.

`macros/materializations/models/table/table_dml_refresh.sql:52-54` — the fused statement:

```jinja
{% call statement('main') -%}
  SELECT * INTO {{ refresh_relation }} FROM {{ tmp_vw_relation }} {{ query_label }}
{%- endcall %}
```

## Cause

Two factors independently hold Sch-M across the load.

**The fused statement.** `SELECT ... INTO` merges DDL and data movement into a single
statement. The Sch-M lock on the new object is taken when the statement begins and released
only when it completes. This holds even under autocommit, since a single statement always
holds its own locks for its full duration.

**The ambient transaction.** `dml_refresh_cleanup_pre` (`table_dml_refresh.sql:41-44`)
issues `DROP TABLE IF EXISTS {{ refresh_relation }}` as the run's first statement. That opens
the ambient dbt transaction and takes Sch-M on the prior scratch table, held until the
materialization's trailing `adapter.commit()` (`table.sql:139`) — spanning the load
independently of the fusion above.

The same fusion appears in `sqlserver__create_table_as` (`macros/relations/table/create.sql:53`),
which serves the rename-swap path, so that path is expected to have the same exposure.

## Impact

A slow model blocks metadata reads for that object across every other session on the
database, for as long as the load runs. The cost of a slow model should fall on the session
running it.

Benjamin-Knight:
Resolve #819 

## Problem

A `table_refresh_method: dml` model held a `Sch-M` lock on its scratch table for the
**whole load**. `Sch-M` is the one lock mode incompatible with the `Sch-S` lock every
metadata reader takes, so while a slow model built, it blocked metadata readers in every
other session on the database — database-wide `sys` / `INFORMATION_SCHEMA` scans, a
concurrent dbt run's catalog and column lookups, SSMS's object explorer. None of those
asked for the scratch table by name.

## Cause

Two independent causes; fixing either alone would have changed nothing.

1. The scratch table was built by one fused `SELECT * INTO`, which holds `Sch-M` on the
   new object from the start of the statement to the end, not for the instant of creation.
2. That build ran inside the materialization's ambient transaction. Locks are held to
   commit, so a split create inside the transaction holds `Sch-M` just as long as the
   fused statement did.

## Fix

- Create the table empty, then load it with a separate `INSERT ... WITH (TABLOCK)` — still
  minimally logged, as `SELECT INTO` was. Hoisted into
  `sqlserver__get_create_table_empty_sql` / `sqlserver__get_tablock_insert_sql` and applied
  to the dml refresh path, the default `table` build, and every other `create_table_as`
  caller (incremental full refresh and temp builds, snapshots).
- Every statement in the build passes `auto_begin=False`, so each autocommits and drops its
  catalog locks as it finishes — the same treatment the incremental temp build already got.
  This also stops the clustered columnstore index built after the load from holding its
  locks to the end of the materialization.
- The dml refresh's `DELETE`+`INSERT` swap commits as soon as it completes, instead of
  running on through index reconciliation, masks, grants and `persist_docs` while holding X
  locks on the target. The swap stays atomic — it is now the whole of its own transaction.

## Worth knowing

- **A pre-hook with `inside_transaction: true` (dbt's default) still re-couples the build.**
  `auto_begin=False` only declines to *open* a transaction; a statement still joins one
  already open. Documented in the macro and the changelog. Same trade-off as the
  incremental path.
- **A crashed run can now leave a `__dbt_tmp` intermediate behind**, since the build commits
  standalone. The existing `OBJECT_ID` guard for adapter-generated throwaways drops it on
  the next run; a fresh create of a real target still surfaces `Msg 2714` rather than
  destroying an object dbt doesn't know about.
- Index and mask reconciliation reconverge on the next run if the swap fails — both
  reconcile against the config rather than applying a delta.
- Snapshots get the split, not the lock change; they keep their own transaction semantics.

## Tests

- New `tests/unit/adapters/mssql/test_table_build_sql.py`: renders the batch that actually
  ships (non-contract, contract, temp build) end to end, so a missing statement terminator
  or an unescaped `EXEC` literal fails without a database. Plus a repo-wide check that no
  macro fuses a create with its load again.
- Re-anchored the query-options hint assertions on the split build. The old swap pattern
  matched the scratch load as readily as the swap, so it would have passed on a log where
  only the swap lost its hint — silently testing nothing. Each pattern is now anchored on
  what its statement selects `FROM`, and verified mutually exclusive against the emitted SQL.


axellpadilla:
Can we move creation of the empty intermediate table before `run_hooks(pre_hooks, inside_transaction=True)` and remove the early `commit_if_open()`?

That keeps the table-creation `Sch-M` outside the model transaction while allowing the long `INSERT` to run without `Sch-M`, preserving the transaction boundary for transactional hooks. The rename/swap can acquire `Sch-M` only at the final cutover.

### Proposed lock/transaction flow

```text
OUTSIDE TRANSACTION
│
├─ outside-tx pre-hooks
│
├─ CREATE empty intermediate table
│      └─ Sch-M
│         short, released immediately
│
▼
BEGIN TRANSACTION
│
├─ inside-tx pre-hooks
│
├─ INSERT INTO intermediate
│      └─ long-running load
│         no Sch-M
│
├─ rename / swap
│      └─ Sch-M acquired here
│
├─ transactional tail
│
└─ COMMIT
       └─ Sch-M released
```

The key property is:

```text
Current problematic shape
begin├──────────────────── long load with Sch-M ────────────────────┤ commit

Proposed
Sch-M ├─ create ─┤
                 begin├──────── long load, no Sch-M ────────┤
                                                       ├─ Sch-M ─┤ commit
                                                          swap
```

One nuance: the `Sch-M` acquired by the rename/swap is held until `COMMIT`, so keeping the post-swap transactional tail short still matters.

Benjamin-Knight:
Not the create — it needs the model SQL to be bindable (the tmp view, or the describe probe on the contract path), so it can't run before a pre-hook that stages what the model reads.

Benjamin-Knight:
Ok, there is a lot of complex changes in the transaction flow here that definitely need another pair of eyes on them @axellpadilla 

I've also added a fix for what I think was a race condition in the usage of the linked server in the open query tests that would intermittently fail my local and the GitHub CI flows.

### Default `table` path (rename swap)

```text
BEGIN  ← first in-tx pre-hook statement, if any
├─ in-tx pre-hooks
│     ▸ scope gate sampled here: transaction_is_open() and scope == 'build'
├─ CREATE VIEW model__dbt_tmp_vw
├─ SELECT TOP 0 * INTO model__dbt_tmp       ← Sch-M, but TOP 0 moves no rows
COMMIT                                       ← Sch-M released, effectively instant
   │
   ├─ INSERT … WITH (TABLOCK)                ← autocommitted; X table lock, never Sch-M
   ├─ CREATE CLUSTERED COLUMNSTORE INDEX     ← autocommitted, on a private name
   │
BEGIN
├─ sp_rename target → backup                 ← Sch-M on the live name
├─ sp_rename intermediate → target
├─ apply_masks                               ← inside: a failure rolls the swap back
├─ in-tx post-hooks
COMMIT                                       ← cutover atomic; Sch-M released
   │
   ├─ create_indexes                         ← the long one, now outside
   ├─ grants / denies / persist_docs
   └─ drop backup, outside-tx post-hooks
```

`INSERT ... WITH (TABLOCK)` takes an exclusive **table** lock, which is compatible with `Sch-S`
— so the long load never blocks a metadata reader. The hint is also what keeps it minimally
logged; it is not there to be removed for "less blocking".

### `table_refresh_method: dml`

```text
   ├─ scratch build (auto_begin=False)        ← create empty, then TABLOCK load
BEGIN
├─ DELETE + INSERT swap                       ← X locks on the target
├─ in-tx post-hooks                           ← now atomic with the swap; previously not
COMMIT
   │
   ├─ reconcile_indexes → apply_masks         ← reconcile first: index drops before re-masking
   ├─ DROP scratch table
   └─ grants / denies / persist_docs
```

The macro no longer commits its own swap. It leaves the transaction open and returns
`schema_match` plus the scratch relation, so the shared tail owns the boundary and picks the
index strategy. On the schema-changed fallback it renames the scratch table into place, masks
inside the transaction, and returns `refresh_relation: none` — that name is already vacated.

### `full_refresh_build: prebuilt`

```text
   ├─ drop existing, rebuild in place         ← marker committed onto its own transaction
├─ apply_masks                                ← inside the load's transaction
├─ in-tx post-hooks
COMMIT
   ├─ create_indexes / grants / denies / persist_docs
```

No swap, so a mask failure leaves an empty target carrying `dbt_full_refresh_incomplete`
rather than rolling back to the old table. That is `prebuilt`'s existing trade, now documented.

### `incremental`

Same gate and same split build on the `create_table_as` branches. Fresh creates now stage
through `__dbt_tmp` and swap — **they previously built straight into the real target, so a
failed first load committed an empty table and the next run merged into it. Silent data
loss.** The append/merge path is untouched: the strategy DML *is* the cutover, so it keeps its
transaction.

---

## What is atomic with what

The transaction spans **in-tx pre-hooks → the cutover → in-tx post-hooks**. That is what a hook
declaring `transaction: true` is asking for: atomicity with *the model*. Index reconciliation,
grants, denies and `persist_docs` are the adapter's own housekeeping and were never part of
that promise.

**Masks are the exception.** On a path that builds a new table it carries no masks until
`apply_masks` runs, so running that after the cutover committed would leave the new table live
with its columns exposed. Masks stay inside the transaction there.

### Two behaviour changes

1. **In-tx post-hooks now run before index creation** (masks are unaffected — they still run
   first). A post-hook that needs indexes present should declare `transaction: false`, which
   runs it after the whole tail.
2. **If you create indexes from post-hooks** — the idiom predating the `indexes` config —
   `drop_unmanaged_indexes: true` now drops them in the same run, and such an index on a masked
   column trips the index-key check on SQL Server before 2022.

## New config

| | Transaction covers | Pre-hook rolls back with a failed load | #819 fixed |
|---|---|---|---|
| `pre_hook_transaction_scope: schema` | pre-hooks + `CREATE VIEW` + empty `CREATE` | no | yes |
| `pre_hook_transaction_scope: build` | pre-hooks + the whole build | yes, except below | no |

`dbt_sqlserver_pre_hook_schema_scope` supplies the default. It ships `False` (meaning `build`)
so current behaviour is preserved, and is expected to flip in a later release.

**Inert where it cannot help:** a model with no transactional pre-hook always takes the fixed
path. On `dml` and `prebuilt` the build joins a pre-hook's open transaction with nothing
committing in between, so `schema` changes nothing there. And `build` cannot deliver rollback
on `prebuilt` or an incremental `--full-refresh` — both commit an in-progress marker before the
load precisely so it survives a failure. `transaction: false` on the hook remains the remedy.

Full detail, including the caveats, in `docs/transaction_scope.md`.

## Also in here

- `auto_begin=False` on seven read-only catalog probes. They were reopening the ambient
  transaction and dragging the tail's mask and index DDL back inside it. A unit test enforces
  the rule so a new probe cannot reintroduce it.
- `sqlserver__create_table_as` split into stage and load halves; `create_table_as` is now their
  concatenation, so snapshots and the incremental temp build emit the same statements in the
  same order. A test pins `whole == stage + load`.
- `adapter.transaction_is_open()` — the config-derived proxy is wrong in both directions
  (`run_hooks` skips hooks whose SQL renders empty; `mark_full_refresh_incomplete` leaves a
  transaction open with no hook involved).
- Podman `make` targets for the local test server, for anyone who would rather not run Docker.
- An unrelated test fix riding along: the openquery tests hard-coded an instance-wide linked
  server name, so concurrent `pytest-xdist` workers dropped and recreated each other's server
  mid-run. Each worker now gets its own.

## Verification

Functional suites against SQL Server 2022 (constraints, hooks, snapshots, grants, masks,
denies, indexes, dml refresh, prebuilt, incremental, concurrency) plus 601 unit tests. The two
`pre_hook_transaction_scope` tests assert **opposite** rollback outcomes and both pass, so the
config demonstrably changes behaviour rather than merely compiling.

axellpadilla:
@Benjamin-Knight Amazing will set time aside to check this in deep

Benjamin-Knight:
pre_hook_transaction_scope is probably one of the larger changes, the intent is that we swap behaviour from current to schema only at some point and users opt in to pre hooks holding the transaction open for the entire build. In most cases we do not want the pre hook to do so and schema is the right call, but there are use cases where the pre hook does need to roll back properly but because of the schema lock this causes users should opt in when they need it rather than cause schema locks for the entire project.

axellpadilla:
I will try some ideas too when checking this, maybe we could use a different connection to create the table at the right time?, I'm currently using latest rc3 with transaction off but I think this PR is the only missing blocker

Feckinotter:
I'm using rc3 with transaction on, other than the schema lock no issues so far. But the schema lock can cause issues, I've also seen an increase in other locks but I'm not sure these are rc3 related as they are parralellism inside queries deadlocking, my only concern is I never had these before transactions in DBT. (Ben on personal account)

axellpadilla:
@Feckinotter yes there should be some limits on parallelism but I'm also not sure that kind of in-process locking have something to do with transactions, here are other best practices that help both with and without transactions https://github.com/dbt-msft/dbt-sqlserver/blob/master/docs/sqlserver-best-practices.md

axellpadilla:
Pushed two commits on top of your branch (e244961 the change, 97b220f a comment pass; plus a merge of current master): schema resolution now runs **before** the in-transaction pre-hooks, which replaces the `schema` scope and the behaviour flag.

**Why.** `schema` committed the pre-hook before the load. That gives the same locks as `transaction: false` on the hook, and once the flag flipped every default pre-hook would have lost its rollback. Staging the create ahead of the hooks keeps rollback and removes the lock window on every path, including the two the branch left open: `dml` with a transactional pre-hook (the issue reporter's steady-state config) and the incremental append temp build. `prebuilt` also no longer reopens the transaction before its load, which held `Sch-M` on the live name through the load regardless of hooks.

**End result, `pre_hook_transaction_scope: load` (default), same shape on `table`, `incremental` and `snapshot`:**

```text
   ├─ outside-tx pre-hooks                       autocommit
   ├─ CREATE OR ALTER VIEW model__dbt_tmp_vw     autocommit
   ├─ SELECT TOP 0 * INTO model__dbt_tmp         autocommit; Sch-M for an instant
   │
BEGIN  ← first in-tx pre-hook statement, if any
├─ in-tx pre-hooks
├─ INSERT INTO model__dbt_tmp WITH (TABLOCK)   ← X table lock, never Sch-M
├─ CREATE CLUSTERED COLUMNSTORE INDEX          ← residual window, documented
├─ sp_rename ×2 / DELETE+INSERT / MERGE        ← the cutover
├─ apply_masks (fresh table only)
├─ in-tx post-hooks
COMMIT                                         ← atomic unit ends; Sch-M released
   │
   ├─ DROP VIEW model__dbt_tmp_vw
   ├─ create_indexes / reconcile_indexes
   ├─ grants / denies / persist_docs
   └─ drop backup, outside-tx post-hooks
```

With no transactional pre-hook, `BEGIN` happens at the cutover and the load and CCI autocommit too.

**Measured on SQL Server 2022** (open transaction in one session, `sys.tables` scan in another): uncommitted `SELECT TOP 0 * INTO`, `CREATE CCI`, `DROP VIEW`, `sp_rename` and `CREATE INDEX` all block the scan; a committed create followed by an uncommitted `INSERT ... WITH (TABLOCK)` does not. That is why the view drop and the index work moved after the commit as well.

**Config.** `pre_hook_transaction_scope: load | build`. `build` keeps today's ordering (create inside the hook's transaction) for the one case `load` cannot serve: a `transaction: true` pre-hook that creates what the model reads, which under `load` fails at the stage with `Invalid object name`. Declaring that hook `transaction: false` also works, since outside-tx hooks run before the stage. Both scopes keep pre-hook rollback, so no behaviour flag is needed; `dbt_sqlserver_pre_hook_schema_scope` and `adapter.transaction_is_open()` are gone.

**Tests.** `test_pre_hook_transaction_scope.py` pins rollback under both scopes, catalog scans unblocked under `load` and blocked under `build` from a second session during an 11s load, and the bindability failure with both remedies. `test_failed_load_leftovers.py` fails a load on the table, dml and snapshot paths and proves the rerun clears `__dbt_tmp`, the tmp views and the scratch table. Unit suite and the materialization functional suites are green. Details in `docs/transaction_scope.md`.

axellpadilla:
@Benjamin-Knight do we really need the third case? (splitting the transaction on build pre_hook_transaction_scope, could be a third option without behavior flag)

Benjamin-Knight:
Doing the schema resolution up front, on its own, is a much better shape than the `schema` scope was, and it does close the paths the branch had left open.

Four things I'd want to sort out first. The first one I think is blocking; the rest are smaller.

## 1. Snapshot pre-hooks now run too late

The two `run_hooks(pre_hooks, …)` calls moved from the top of the snapshot materialization down to `snapshot.sql:65` and `:77`. That leaves three things that talk to the database sitting *above* them now:

```
master                              now
--------------------------          --------------------------
pre-hooks                           work out the strategy   :26   ← reads the DB
work out the strategy               check the target        :55
check the target                    drop the staging table  :59
                                    pre-hooks               :65, :77
```

The one that bites is `snapshot.sql:26`. On a `check`-strategy snapshot, dbt-core runs the snapshot's own SQL there to see which columns it returns. So if a pre-hook is what creates the thing the snapshot reads, that SQL can't bind — and because dbt skips the column check entirely while the snapshot table doesn't exist yet, it passes on the first run and only fails on the second:

```
run 1   no snapshot table yet  → column check skipped   → PASS
run 2   snapshot table exists  → runs the model SQL     → Invalid object name
                                 (pre-hooks still 39 lines away)
```

Neither escape hatch reaches it: `build` only moves the stage from `:71` to `:82`, still below `:26`, and `transaction: false` doesn't help either now that `:65` is below it too. So on these snapshots the new config doesn't actually give you a way out, and it's a regression against master rather than just against the branch. Moving both `run_hooks` calls back above `:25` should be all it takes — the stage can stay where it is.

## 2. `build` takes the lock even when nothing asked it to

`build` is meant to be "the old behaviour, and only for models that opt in". The old gate checked whether a transaction was actually open before doing anything; that check went away with `transaction_is_open` and nothing replaced it. So the build statement now opens a transaction *itself* when no pre-hook did — `table.sql:144`, `incremental.sql:134`, `snapshot.sql:85`.

That matters because a transaction holds its locks until it commits. Create the table inside one and the schema lock — the one that blocks anything reading table metadata anywhere on the database — stays up for the whole load instead of an instant:

```
what the docs promise              what actually happens
------------------------           ------------------------
create table    autocommit         BEGIN
load            autocommit         |- create table   ← lock taken here
BEGIN                              |- load           ← ...still held
|- rename, masks                   |- rename, masks
COMMIT                             COMMIT            ← released only here
```

`docs/transaction_scope.md:203-205` says this outright: "With no transactional pre-hook the setting changes nothing." It isn't a regression against master — master held the lock there too — but a folder-level `+pre_hook_transaction_scope: build`, which is what the docs suggest, would quietly re-open #819 for every model underneath it that has no hooks at all. Looks like a one-line `auto_begin=False` in each of the three spots; the cutover still opens its own transaction afterwards either way.

## 3. "Both scopes keep pre-hook rollback" is a bit stronger than the code

Two paths commit the pre-hook before the load no matter what the scope says, because they write an "in progress" marker first and that marker has to survive a failure:

- `full_refresh_build: prebuilt` — `create.sql:302`
- incremental `--full-refresh` of an existing table — `incremental.sql:117`

Once that commits, the hook's writes are durable and there's nothing left to roll back. This is already caveated correctly in the macro comments, the changelog and `docs/transaction_scope.md:193-201` — it's just the PR comment and the README that state it flat. Worth tightening, since unconditional rollback is the stated reason we didn't need a behaviour flag. With the caveat added the argument still holds: rollback is kept everywhere the flag would have mattered.

## 4. The test description doesn't match the tests any more

The write-up describes catalog scans blocked/unblocked from a second session; `4378e63` replaced that with reading the lock DMVs on the building session. I think that's the better instrument for the reason the commit message gives, and it isn't a test that can quietly pass on nothing. Two things though:

- It skips silently when the login lacks `VIEW SERVER STATE` (`:225-228`). If CI doesn't run as sysadmin, the only functional guard for #819 vanishes with no signal. Could that be a hard failure instead?
- `:265` accepts the lock being held for half the samples, but the claim is that it's held the whole time. Asserting all of them would actually pin the mechanism.

## Smaller things

- `load` doesn't quite remove the window on *every* path — the dml schema-change rebuild still runs inside the cutover transaction (`table_dml_refresh.sql:127` and `:165`). That predates this PR and I'd leave the code alone, but it's a `dml` model, which is the reporter's setup, so it's worth calling a known residual rather than covered.
- Staging early fixes the intermediate's *column shape*, not just its existence. A pre-hook that widens or adds a column to a source now shows up as a truncation error during the load rather than a clear "invalid object name" at the stage. Same cause, much less obvious message — worth a sentence next to the existing caveat in the docs.
- The new `DROP VIEW`s on the tail (`table.sql:187,191`, `incremental.sql:219`, `snapshot.sql:166`) don't carry a `USE`, where every other drop in the adapter does. Fine today, but on a cross-database model it could no-op against the wrong database and leave a stray view behind.

## On your question about the third case

> do we really need the third case?

If you mean bringing `schema` back as a third value — I don't think so. It gives the same locks as `load` and loses the rollback, and the one thing it does that `load` doesn't (commit the hook before the load) is exactly what `transaction: false` on the hook already means. It'd be a second way to say the same thing.

If you mean `build` itself — I'd keep it. It's the only setting that gives you both bindability *and* rollback, which is the case that killed `schema`: a hook that has to create what the model reads *and* be undone if the load fails. `transaction: false` buys the first by giving up the second.

But it's narrower today than we're claiming: inert on `prebuilt`, no rollback on incremental `--full-refresh`, and on `check`-strategy snapshots it doesn't give bindability either (point 1) — so there it currently buys nothing over `load`. Fixing point 1 is what makes "keep it" a real answer rather than a nominal one. Two values feels right to me once 1 and 2 are in.

axellpadilla:
Merged `master` in to clear the conflicts (`e1edcdc`). Everything that collided came from #828 (constraints in contract-enforced builds), which landed after this branch was cut.

**`tests/functional/adapter/dbt/test_constraints.py`** — the interesting one. Neither side's `expected_sql` was correct after the merge:

- this branch splits the `CREATE TABLE` and the `INSERT ... WITH (TABLOCK)` into separate `EXEC` batches, but predates constraints being emitted at all;
- master carries the `check` constraints but still has both statements fused into one `EXEC`.

The merged macros emit **both** — a `CREATE TABLE` carrying the constraints, then a separate `EXEC` for the TABLOCK insert — which is exactly what you want from the merge. Verified against the real generated artifact rather than hand-written:

```sql
EXEC('CREATE TABLE <model_identifier>
    ( id INT not null, color varchar(100), date_day varchar(100),
      check ((id > 0)), check (id >= 1) )
')
EXEC('INSERT INTO <model_identifier> WITH (TABLOCK) ("id", "color", "date_day")
        SELECT "id", "color", "date_day" FROM <model_identifier> ')
```

Both fixtures were regenerated from that.

**`CHANGELOG.md`** — both sides had rewritten the `table_refresh_method: dml` columnstore entry. Kept this branch's mechanics, since they are what is true after the merge (empty create plus a separate `INSERT ... WITH (TABLOCK)`, not a fused `SELECT * INTO`), and took master's wider scope, since after #828 that path loses declared constraints as well as `NOT NULL`s. All other entries from both sides kept as they were.

**No macro conflicts.** `columns_spec_ddl.sql` merged identical to master's, and the split create path in `create.sql` calls `build_columns_constraints`, so constraints survive the split — that is why only the fixtures needed touching.

Full suite on SQL Server 2022 after the merge:

- `tests/functional` — **401 passed, 48 skipped, 2 xfailed** (`-n 8`)
- `tests/unit` — **628 passed, 5 skipped**

PR is mergeable again. The other open PRs (#841, #842) both merge cleanly on top of this, so this can go first as intended.

PR #829, opened 2026-08-30
fix: rebuild the DML refresh's scratch table so a schema change keeps the columnstore

lll86789:
Split out of #828 at review request, so it can land independently of the constraints work.

## The bug

A model with `table_refresh_method: dml` loses its clustered columnstore index the first time its schema changes, and never gets it back.

That path materializes a scratch table with `SELECT * INTO` and compares its columns against the target. When they match it refreshes by `DELETE` + `INSERT`, which leaves the target object — and its indexes — in place. When they don't, it renames the scratch into position instead.

`SELECT * INTO` copies no index and no constraint, and infers nullability from the query. `create_indexes` then builds only what the `indexes` config names — never the `as_columnstore` CCI, which the table materialization builds separately. So after the rename the model is a heap.

It stays one. Every later run now matches the new schema, takes the `DELETE`+`INSERT` path, and never renames again — so nothing ever rebuilds the index. The failure is silent: every run reports success, and the only symptom is that queries got slower.

Under an enforced contract the same rename also drops the model's `NOT NULL`s, since the scratch took its nullability from the query rather than from the contract.

## The fix

On the rename branch — and only there — the scratch is rebuilt through `create_table_as` before the swap. That is how every other build path in this adapter creates a table, so the columnstore index and the full column DDL carry across.

`SELECT * INTO` stays as the schema probe. Doing the rebuild up front would build, and then throw away, a full columnstore index on every steady-state refresh, which on a large table dominates the run. A schema change is rare, so one extra build there is much the cheaper trade.

## Cost

On a schema-change run, and only there:

- The model's SQL executes twice — once for the `SELECT … INTO` probe, once for the rebuild.
- One extra columnstore build.
- The run gives up minimally-logged `SELECT … INTO`. `INSERT … WITH (TABLOCK)` is itself minimally logged under the simple and bulk-logged recovery models, so the extra log volume lands on full-recovery databases only.

Steady-state refreshes are unchanged.

Probing the tmp view rather than the materialized scratch would avoid the double execution, but it changes how the probe behaves and belongs in its own change.

## Tests

`TestDmlRefreshColumnstoreSurvivesSchemaChange` in `tests/functional/adapter/mssql/test_table_refresh_method.py`: build a `table_refresh_method: dml` model on the default `as_columnstore`, add a column to force the rename-swap fallback, assert the new column arrived and the CCI is still there. Deliberately contract-free — the loss is not contract-specific.

Run against SQL Server 2022 (the repo's `docker compose` image), `mssql-python` backend:

| | |
| --- | --- |
| The new test, with this branch's macro reverted to `master`'s | **fails** — `AssertionError: assert False` on `assert has_columnstore_index(project, "dml_cci_swap_model")` after the column is added |
| The new test, on this branch | **passes** |
| `test_table_refresh_method.py` in full, on this branch | **14 passed** |

The failing run is worth a look, because it is what the bug looks like in practice: dbt reports `Completed successfully` / `PASS=1` and logs `Schema change detected … falling back to rename-swap` exactly as expected. The first run's columnstore assertion passes; only the one after the schema change fails. Nothing in the run output says anything is wrong.

## Relationship to #828

#828 is rebased on this branch and keeps the constraint-specific half: `TestDmlRefreshKeepsConstraints`, which asserts the named `PRIMARY KEY`, the inline `CHECK` and the `NOT NULL`s all survive the same swap. That test only means anything once #828's constraint emission is in, which is why it stays there.

The dependency runs one way. This change calls `get_create_table_as_sql` and touches none of #828's new code; #828 needs this change, because without it the rename-swap washes away the constraints it just built.


Benjamin-Knight:
Looks good, although we have a test to cover the CCI part of this bug and not the constraint part of it which is important as its the basis of #828.

lll86789:
Agreed — the NOT NULL half is testable here on its own, since render_column_constraint already emits not null on master and SELECT * INTO is what drops it. I'll add a test that puts a contract on a dml model, forces the schema-change fallback, and reads IS_NULLABLE back out afterwards.

lll86789:
Done. I added a test that puts a contract on a dml model, forces the schema change, and checks the NOT NULLs are still there afterwards. Confirmed it fails without the fix.
PK and CHECK will have to wait for #828, they're not emitted at all until that lands.
Also snuck in a third commit fixing my own changelog wording — I'd got the rebuild's bulk-load shape wrong, and it didn't mention that already-broken tables need a --full-refresh to come back.

Benjamin-Knight:
Those failures are a functional tests issue, the fix is in another branch right now wrapped in some transaction work.

https://github.com/dbt-msft/dbt-sqlserver/pull/822/changes/b57204ca4ef1ec4fb074d2b31bdf70d5c90b2d0a

You will either need to wait for that branch to land and or cherry pick out that commit into this branch as well to get the tests passing.

axellpadilla:
Looks good overall. The added CCI and NOT NULL regression coverage gives me confidence in the schema-change fix.

One thing before merging: I think the new recovery note for already-affected tables is incorrect for the default DML configuration. With `table_refresh_method: dml` and the default `full_refresh_build: heap_then_index`, `--full-refresh` still takes the DML path. If the damaged heap already matches the model schema, that path does DELETE + INSERT followed by index reconciliation, which does not recreate a missing implicit `as_columnstore` CCI.

`full_refresh_build: prebuilt` together with `--full-refresh` would force the actual rebuild and restore the CCI, but plain `--full-refresh` with the default settings does not.

I’ll also backport this fix to `v1.11.x` because this deserves it. Please cherry-pick that commit if you can; if not, I’ll work on a couple of things and review today anyway.

lll86789:
@axellpadilla You're right! I'll update the recovery note to require full_refresh_build: prebuilt with --full-refresh. Thanks for taking the v1.11.x backport.

@Benjamin-Knight I'll cherry-pick the test fix from #822. Thanks!

PR #845, opened 2026-09-18
fix(describe): keep the query's own error instead of retrying it blind

### LINKED ISSUE #844: Invalid column name surfaces as "Attempt to use a closed connection"
A model selecting a column that does not exist upstream fails with:

    Database Error in model {model_name}
      Attempt to use a closed connection.

The actual error reaches `logs/dbt.log` at debug level only:

    Invalid column name {name}. (207)

## Cause

`_describe_result_set` catches every exception from
`add_select_query(describe_sql)` and returns `None`, sending
`get_column_schema_from_query` on to execute the query. Msg 207 is not the
procedure declining to describe — the query does not compile, so executing
raises the same thing. By then the error handler has rolled back and closed the
connection, so the fallback fails on that instead and the transport message
replaces the real one.

## Scope

Any CTE-headed query reaching `get_column_schema_from_query` — contract-enforced
models, snapshot staging — on any materialization. Seen on `table` and on
incremental `full_refresh_build=prebuilt`.

Retries don't help; a missing column is not transient. A scheduled run fails
twice and the task log gives an on-call reader nothing to act on.

## Repro

Any model with a CTE at the head whose SELECT names a column its upstream lacks.

Benjamin-Knight:
`_describe_result_set` treated every describe failure as the procedure declining
to describe a good query, and fell back to executing.

Classify instead: `sp_describe_first_result_set` reports what it cannot describe
with its own 11500-11599 family (11514 `#temp`, 11526 uncompilable), which every
backend carries through in the message text. Those keep the fallback. Anything
else is raised — executing raises it identically, and no fallback can produce
column metadata for a query that does not compile.

Fixes #844 .

## Tests

Unit coverage for both directions in `test_get_column_schema_from_query.py`
(207/208 raise without a second `add_select_query`; 11514/11526 still execute).
`test_describe_error_surfaces.py` builds a contract-enforced CTE model naming a
column its upstream lacks and asserts the output carries the column name and not
`closed connection` — on master that test reports `Cannot create cursor on
closed connection`.


axellpadilla:
The 207/208 propagation looks correct, but I think the 115xx fallback still has a connection lifecycle issue: `add_select_query()` hits `exception_handler()`, which calls `release()` before re-raising. So after a real 11514/11526, `_describe_result_set()` returns `None` and the fallback immediately calls `add_select_query(sql)` again on the same now-closed thread connection.

The current unit test mocks `add_select_query()` directly, so it skips that lifecycle. Could we add an integration test with a real describe failure, for example a query where `EXEC sys.sp_describe_first_result_set N'...'` returns 11514/11526, and verify the fallback actually succeeds?

Also, `sys.dm_exec_describe_first_result_set` instead of calling sp_describe_first_result_set exposes structured `error_number/error_type` metadata and could avoid both the connection-release problem and parsing `str(e)` with a regex.

Issue #838, opened 2026-09-09
Child views show stale data types when parent view is updated


csaphn-jkaczanowski:
Core:
  - installed: 1.12.0
  - latest:    1.12.4 - Update available!

  Your version of dbt-core is out of date!
  You can find instructions for upgrading here:
  https://docs.getdbt.com/docs/installation

Plugins:
  - sqlserver: 1.11.1 - Up to date!

**steps to reproduce**

1. create following views

vw_parent.sql
```
{{
  config(
    materialized='view',
  )
}}

select '1' as my_col
```

vw_child.sql
```
{{
  config(
    materialized='view',
  )
}}

select my_col from {{ ref('vw_parent') }}
```

2. dbt run

3. modify vw_parent.sql (change the varchar column to int)
```
{{
  config(
    materialized='view',
  )
}}

select 1 as my_col
```

4. dbt run

**expected output**
vw_child.my_col is type int

**actual output**
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:w="urn:schemas-microsoft-com:office:word"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns:m="http://schemas.microsoft.com/office/2004/12/omml"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Word.Document>
<meta name=Generator content="Microsoft Word 15">
<meta name=Originator content="Microsoft Word 15">
<link rel=File-List
href="file:///C:/Users/JKACZA~1/AppData/Local/Temp/72/msohtmlclip1/01/clip_filelist.xml">

<link rel=themeData
href="file:///C:/Users/JKACZA~1/AppData/Local/Temp/72/msohtmlclip1/01/clip_themedata.thmx">
<link rel=colorSchemeMapping
href="file:///C:/Users/JKACZA~1/AppData/Local/Temp/72/msohtmlclip1/01/clip_colorschememapping.xml">

<style>

</style>

</head>

<body lang=EN-AU style='tab-interval:36.0pt;word-wrap:break-word'>

TABLE_SCHEMA | TABLE_NAME | COLUMN_NAME | ORDINAL_POSITION | COLUMN_DEFAULT | IS_NULLABLE | DATA_TYPE
-- | -- | -- | -- | -- | -- | --
dbo | vw_child | my_col | 1 | NULL | NO | varchar

</body>

</html>

[...]

PR #847, opened 2026-09-19
fix(view): refresh a skipped view's metadata only when it is stale

axellpadilla:
Improves the #838 fix. That fix ran `sp_refreshview` on every skipped view, which advances `sys.objects.modify_date` exactly as an `ALTER` does — so every unchanged view in a project churned that column on every run, and it could no longer tell a skipped run from a rebuild. Anything watching the catalog for drift saw noise, and the test that used to assert "no rebuild" had to give up its signal.

## What changed

The skip decision already fetched the stored definition. It now gets the view's cached column shape in the same query, compares it against what the body resolves to (`sys.dm_exec_describe_first_result_set`, which compiles the body without executing it), and runs `sp_refreshview` only where the two disagree.

- Same round trip as before, one compile of the body per unchanged view.
- A view whose sources have not moved is left completely untouched, so `modify_date` works as a signal again.
- Fails toward refreshing: a body that cannot be described is refreshed rather than assumed current.

The check is a **new macro**, `get_view_skip_check_sql`, rather than a second column on `get_view_definition_sql` — a project overriding that macro would otherwise silently lose a column the materialization then indexes, on the path that runs for every unchanged view.

The text comparison is unchanged and still the gate. A shape check cannot replace it: `where amount > 100` → `> 200` resolves to an identical column list, so a probe-only skip test would skip a real edit — the failure mode #807 fixed.

## Notes

- Both `sysname` comparisons are pinned with `COLLATE DATABASE_DEFAULT`. Without it every view model fails with *Cannot resolve the collation conflict* wherever the catalog and database collations differ — caught by the functional suite, not by review.
- `is_nullable` is deliberately not compared: describe's inference over expressions can disagree with what `CREATE VIEW` recorded, which would refresh that view on every run forever.
- `sys.columns` is left unhinted, per the existing policy for catalog reads that act as correctness guards.

## Testing

Three test classes collapse into one — [`TestSkippedViewTracksItsSourceShape`](tests/functional/adapter/mssql/test_materialize_change.py), one project, one method walking a single view's life: build → quiet run → added column → retyped column → quiet run. It pins both directions, including that a repaired view goes quiet again. `TestViewMaterializationNoOp` gets its `modify_date` assertion back.

Verified against SQL Server 2022: 646 passed, 5 skipped.

Targets the 1.12.0rc4 release being prepared in #841. The 1.11 backport branch (`backport/1.11-838-view-refreshview`) carries the same `sp_refreshview` behaviour and will need this separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


joshmarkovic (approved):

joshmarkovic:
Excellent!! I'll test out the new releases today :)

PR #850, opened 2026-09-21
fix(indexes): use logical relation for CCI naming and clean up legacy names

mopthe:
## Summary

Fix `as_columnstore` CCI naming so creates and rebuilds no longer keep `__dbt_tmp` in the persisted CCI name after swap. This left a temporary-looking index name on the final table.

This change makes the index name follow the logical/final relation while keeping the physical relation as the actual target for `CREATE INDEX`, and it cleans up legacy `__dbt_tmp_cci` names before create.

## What changed

- `sqlserver__create_clustered_columnstore_index` now uses the logical/final relation for the persisted CCI name
- the physical relation remains the object used for `CREATE INDEX`
- `sqlserver__get_create_table_load_sql` now takes `logical_relation` from all callers (`table`, `incremental`, `snapshot`)
- legacy `__dbt_tmp_cci` names are dropped before create to preserve compatibility without forcing a full-refresh upgrade
- added regression coverage for naming behavior and legacy cleanup

## Why

After a swap/rebuild, the persisted CCI name could still carry a temporary relation suffix, which made the final table look like it was still using a transient object name. This patch keeps the persisted name stable while still handling the physical target correctly for SQL Server DDL.

## Testing

- `uv run pytest test_index_configs.py -q`

This validates the `as_columnstore` CCI naming fix and covers the naming regression described in  [#578](https://github.com/dbt-msft/dbt-sqlserver/issues/578), without claiming to address every aspect of the broader issue.


dbt/include/sqlserver/macros/adapters/indexes.sql
```diff
@@ -21,9 +25,24 @@
         AND object_id=object_id('{{ escape_single_quotes(relation_name) }}')
     )
     DROP index {{ relation_name }}.{{ adapter.quote(cci_name) }}
+    {#- Backward-compatibility for old CCI names created before the logical/physical
```
Benjamin-Knight: This section is trying to delete the CCI from relation_name which is the temporary object, not the old object with the old CCI name on it.

tests/functional/adapter/mssql/test_full_refresh_build.py
```diff
@@ -359,6 +365,36 @@ def test_cci_prebuilt_lifecycle(self, project, unique_schema):
         )
         assert leftovers[0] == 0
 
+    def test_legacy_cci_name_is_replaced_on_subsequent_full_refresh(self, project, unique_schema):
```
Benjamin-Knight: Does this test actually test anything? This would pass on the current code because it uses the prebuilt path so the indices have the same name and the bit under test is skipped?

dbt/include/sqlserver/macros/adapters/indexes.sql
Benjamin-Knight: The usage of the jinja white space collapsing here generates unusual looking SQL code, I think it may work but it would be better if we preserved the white space so that its both readable and will work without breaking in an unusual way further down the line.

Benjamin-Knight (changes requested):

mopthe:
@Benjamin-Knight Thanks!
1. Legacy CCI cleanup now drops on the physical relation (the final model table), not the logical temp build table.
2. I removed the full-refresh test you called out, since it didn’t exercise the logical/physical split path.
3. I updated the Jinja whitespace handling to keep the rendered SQL readable and consistent with the style in master.

axellpadilla:
Rebased on master and reworked, addressing the review:

- Legacy cleanup removed. It dropped the CCI on the live table mid-build (a full rewrite to heap right before the swap discards it), and it's unneeded: the CCI is only ever created on a freshly built table, so an old name can't collide.
- The prebuilt test additions are replaced by `test_cci_name.py`, which covers the swap paths (table, incremental, snapshot) and fails on master.
- The Jinja-patching unit test is dropped; the CHANGELOG edit is now `changes/578.bugfix.md`.

@mopthe your commits are kept; the rework is in f537f07.

